Issue #93, opened 2014-04-14
Feature request: download button


Gnewbee:
Please include a download button for git illiterates to be able to download/use the code.

PR #94, opened 2014-04-15
Added download option to the Tree view

lb1a:
Using the dulwich archive function combined with a temporary file and
Flask builtin send_file function.

Adressed to Feature Request #93


jelmer:
FWIW recent Dulwich has a porcelain.archive() function that can generate an archive.

jonashaag:
Thanks @jelmer, didn't know that. Maybe we can replace our custom archiving code with that at some point, although it's not a priority for me right now.

PR #107, opened 2015-01-19
add option to auto-launch browser on server

rjw57:
Add command-line argument --browser (shortened to -b) which will
auto-launch a web-browser pointing to the Klaus server via the Python
webbrowser module.

Note that we do things in the wrong order in that we launch the browser
and *then* start the server. In the absence of a cleaner way of getting
a function to run after server setup this approach actually works in
practice. If this doesn't Just Work for someone we can think again about
more complex options.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 0.4.9 (TBA)
 -----------
 * Bug #104: "git" executable unnecessarily required to be available (@Mechazawa)
+* Add option to auto-launch a web-browser on startup (@rjw57)
 
 0.4.8 (June 22, 2014)
 ---------------------
--- a/bin/klaus
+++ b/bin/klaus
@@ -4,6 +4,7 @@
 import sys
 import os
 import argparse
+from webbrowser import open as open_url
 
 from dulwich.errors import NotGitRepository
 from dulwich.repo import Repo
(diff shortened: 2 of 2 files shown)


jonashaag:
Nice patch, but could you tell me a bit more about your use case? Couldn't you just create a small script that does this?

rjw57:
My workflow is "I want git instaweb which looks nice".

I *could* write a wrapper script. That's true of many features though, hence why git itself is wrapper scripts all the way down :). But consider: a) the script would need to be different on, .e.g, OS X and Linux, it's better to use the experience baked into the webbrowser module [*] b) writing a script doesn't Just Work, it requires everyone wanting to replace "git instaweb" to write a wrapper script[**], c) how does the wrapper script know the URL of the server without hard-coding the host and port or overriding both and d) it makes klaus a full replacement for the git instaweb workflow.

My particular use case was d). git instaweb is really useful for "quickly browse this repo" but gitweb is relatively unattractive and feature poor. Having a nicer alternative a pip-install away is compelling. Having that alternative a pip-install-and-then-write-a-wrapper less so. So this PR is less for me now I've made the change but and more for everyone else like me who Googled for instaweb replacements.

Again, one *could* write a wrapper script. But then one *could* do, e.g., host and port configuration through environment variables and then say that users who want a command-line option write a wrapper script which sets the appropriate variable. Whether something is worthy of wrapping or being rolled into the main utility is a taste judgement which I'm happy to defer to the project's maintainer.

[*] Of course one could write the wrapper script in Python.
[**] So, even if this doesn't become part of the utility, shipping an example wrapper script with klaus would be good documentation at least.

jonashaag:
This convinces me. Thanks for the patch. I'm going to make a few changes to make it a bit more flexible.

rjw57:
Super, thanks. There's a small problem with your modification though, see #108. I'm not sure how you'd want this fixed which is why I didn't fix it myself.

PR #109, opened 2015-01-20
split browser option into flag and option

### LINKED ISSUE #108: --browser option broken with no argument
In PR #107, the --browser (-b) option was added. Commit fab7f84d added an optional argument to the parameter but the parsing is broken. Specifically:

```console
$ klaus -b .
```
which is intended to open klaus browsing the current directory in the default browser instead interprets "." as the browser to launch. This manifests as the rather unobvious error reproduced below:
```
$ klaus -b .
WARNING: No repositories supplied -- syntax is 'klaus dir1 dir2...'.
  File "/home/rjw57/.local/bin/klaus", line 10, in <module>
    execfile(__file__)
  File "/home/rjw57/Downloads/klaus/bin/klaus", line 86, in <module>
    main()
  File "/home/rjw57/Downloads/klaus/bin/klaus", line 67, in main
    _open_browser(args)
  File "/home/rjw57/Downloads/klaus/bin/klaus", line 81, in _open_browser
    opener = webbrowser.get(args.browser).open
  File "/usr/lib64/python2.7/webbrowser.py", line 52, in get
    raise Error("could not locate runnable browser")
webbrowser.Error: could not locate runnable browser
```
Perhaps this can be improved by splitting --browser into a separate flag and option? Or dropping the argument? I'd create a PR but I don't know which approach @jonashaag would favour. As it stands, though, the feature is broken.

rjw57:
As noted in #108, the ``--browser`` option is currently broken if no
argument is supplied. Split the argument into two: a flag
(``--browser``) indicating if a browser is to be opened and an option
(``--with-browser``) which may optionally be used to specify which
browser is to be used.

Fixes #108.

### DIFF
--- a/bin/klaus
+++ b/bin/klaus
@@ -29,8 +29,10 @@ def make_parser():
     parser.add_argument('--host',     help="default: 127.0.0.1", default='127.0.0.1')
     parser.add_argument('--port',     help="default: 8080", default=8080, type=int)
     parser.add_argument('--site-name', help="site name showed in header. default: your hostname")
-    parser.add_argument('-b', '--browser', help="open klaus in default browser on server start. Optional argument: browser executable to use",
-                        default=None, const='__default_browser__', nargs='?')
+    parser.add_argument('-b', '--browser', help="open klaus in a browser on server start",
+                        default=False, action='store_true')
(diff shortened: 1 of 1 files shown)


rjw57:
I could also move the two arguments to a separate group if wanted.

jonashaag:
Nice! Thanks!

PR #110, opened 2015-01-20
default to current directory as repo if none specified

rjw57:
This is a pure laziness PR :). It's nice when cloning a new repo to poke around. Currently, the quickest way to pop up a klaus view of a repo you're already in is:

```console
$ klaus -b .
```

This PR adds a single commit which just changes the (currently unassigned) "no repos specified" behaviour to use the current working directory. My use case for this is extremely lazy: I don't want to type "space" and ".". It does follow the usual git "implicitly this repo" rules though.

Note: no attempt is made to determine the repo root. Running this from somewhere deep in the repo won't work. It could be made to work but it's probably overkill.

### DIFF
--- a/bin/klaus
+++ b/bin/klaus
@@ -32,7 +32,7 @@ def make_parser():
     parser.add_argument('-b', '--browser', help="open klaus in default browser on server start. Optional argument: browser executable to use",
                         default=None, const='__default_browser__', nargs='?')
 
-    parser.add_argument('repos', help='repositories to serve',
+    parser.add_argument('repos', help='repositories to serve or, if omitted, use current directory',
                         metavar='DIR', nargs='*', type=git_repository)
 
     grp = parser.add_argument_group("Git Smart HTTP")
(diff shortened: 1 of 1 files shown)


jonashaag:
Requiring that repository paths be specified explicitly is a design decision:

- It makes very obvious that klaus makes no attempt to look for repositories itself in a folder etc, let alone update the list of hosted repositories during runtime.
- It avoids security issues due to misconfiguration, e.g. when accidentally starting from the wrong directory or getting the shell argument list wrong:

  Consider some script that takes as input a list of repositories to host and passes that list to klaus. Now for some reason, that list may be empty, which hasn't been taken care of by the author of the script. Suddenly we have a possible security problem here, because now klaus hosts the repository found in the current working directory, which might not be intended.

rjw57:
Righty-ho. This was just motivated by laziness.

rjw57:
Closing because I'm retracting this PR.

jonashaag:
> This was just motivated by laziness.

Which is usually a very good reason for filing a bug or coding something in the first place :-)

PR #112, opened 2015-02-01
Added a wsgi application that automatically discovers all git repositori...

ldrumm:
...es in KLAUS_REPOS. `klaus.contrib.autodiscover:application`

Also watches for changes to KLAUS_REPOS using the most efficient filesystem notifier available on the system (pip install watchdog)

I made slight changes to `klaus.Klaus` to enable the integration with the filesystem watcher, but these changes should be backwards-compatible.
Usage of the new autodiscover module requires [`watchdog`](https://pypi.python.org/pypi/watchdog/)

This new wsgi application significantly eases administrative burden when creating/deleting/moving git repositories.  The web server no-longer requires a restart.

### DIFF
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -16,7 +16,7 @@ class Klaus(flask.Flask):
     }
 
     def __init__(self, repo_paths, site_name, use_smarthttp, **kwargs):
-        self.repo_map = dict((repo.path, repo) for repo in map(FancyRepo, repo_paths))
+        self.repo_map = dict((path, FancyRepo(path)) for path in repo_paths)
         self.site_name = site_name
         self.use_smarthttp = use_smarthttp
 
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -15,18 +15,14 @@ class Klaus(flask.Flask):
         'undefined': jinja2.StrictUndefined
     }
 
-    def __init__(self, repo_paths, site_name, use_smarthttp):
-        self.repos = map(FancyRepo, repo_paths)
-        self.update_repos_list()
+    def __init__(self, repo_paths, site_name, use_smarthttp, **kwargs):
+        self.repo_map = dict((repo.path, repo) for repo in map(FancyRepo, repo_paths))
         self.site_name = site_name
         self.use_smarthttp = use_smarthttp
 
         flask.Flask.__init__(self, __name__)
 
         self.setup_routes()
-
-    def update_repos_list(self):
-        self.repo_map = dict((repo.name, repo) for repo in self.repos)
 
     def create_jinja_environment(self):
         """ Called by Flask.__init__ """
@@ -64,7 +60,7 @@ class Klaus(flask.Flask):
             self.add_url_rule(rule, view_func=getattr(views, endpoint))
 
 
-def make_app(repos, site_name, use_smarthttp=False, htdigest_file=None, klass=Klaus):
+def make_app(repos, site_name, use_smarthttp=False, htdigest_file=None, klass=Klaus, **kwargs):
     """
     Returns a WSGI app with all the features (smarthttp, authentication)
     already patched in.
@@ -73,13 +69,14 @@ def make_app(repos, site_name, use_smarthttp=False, htdigest_file=None, klass=Kl
         repos,
         site_name,
         use_smarthttp,
+        **kwargs
     )
     app.wsgi_app = utils.SubUri(app.wsgi_app)
 
     if use_smarthttp:
         # `path -> Repo` mapping for Dulwich's web support
         dulwich_backend = dulwich.server.DictBackend(
-            dict(('/'+repo.name, repo) for repo in app.repos)
+            dict(('/'+repo.name, repo) for repo in app.repo_map.itervalues())
         )
         # Dulwich takes care of all Git related requests/URLs
         # and passes through everything else to klaus
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -17,13 +17,16 @@ class Klaus(flask.Flask):
 
     def __init__(self, repo_paths, site_name, use_smarthttp):
         self.repos = map(FancyRepo, repo_paths)
-        self.repo_map = dict((repo.name, repo) for repo in self.repos)
+        self.update_repos_list()
         self.site_name = site_name
         self.use_smarthttp = use_smarthttp
 
         flask.Flask.__init__(self, __name__)
 
         self.setup_routes()
+
+    def update_repos_list(self):
+        self.repo_map = dict((repo.name, repo) for repo in self.repos)
 
     def create_jinja_environment(self):
         """ Called by Flask.__init__ """
@@ -61,12 +64,12 @@ class Klaus(flask.Flask):
             self.add_url_rule(rule, view_func=getattr(views, endpoint))
 
 
-def make_app(repos, site_name, use_smarthttp=False, htdigest_file=None):
+def make_app(repos, site_name, use_smarthttp=False, htdigest_file=None, klass=Klaus):
     """
     Returns a WSGI app with all the features (smarthttp, authentication)
     already patched in.
     """
-    app = Klaus(
+    app = klass(
         repos,
         site_name,
         use_smarthttp,
--- a/klaus/contrib/autodiscover.py
+++ b/klaus/contrib/autodiscover.py
@@ -1,7 +1,6 @@
 import os
 import dulwich
 import logging
-import functools
 
 from klaus.repo import FancyRepo
 from klaus import Klaus, make_app
@@ -12,7 +11,6 @@ try:
         FileSystemEventHandler,
         DirCreatedEvent,
         DirDeletedEvent,
-        DirMovedEvent
     )
 except ImportError:
     import sys
@@ -119,7 +117,7 @@ class KlausAutoDiscover(Klaus):
     """
     Instead of the normal Klaus behaviour of receiving a list of git repositories
     as `os.environ['KLAUS_REPOS']`, this app searches `os.environ['KLAUS_REPOS']`
-    for valid git repositories and updates the application automatically when any 
+    for valid git repositories and updates the application automatically when any
     changes are made.
     This simplifies administration significantly, but may cause problems with
     large repositories.
--- a/klaus/contrib/autodiscover.py
+++ b/klaus/contrib/autodiscover.py
@@ -1,0 +1,167 @@
+import os
+import dulwich
+import logging
+import functools
+
+from klaus.repo import FancyRepo
+from klaus import Klaus, make_app
+
+try:
+    from watchdog.observers import Observer
+    from watchdog.events import (
+        FileSystemEventHandler,
+        DirCreatedEvent,
+        DirDeletedEvent,
+        DirMovedEvent
+    )
+except ImportError:
+    import sys
+    print("To automatically detect and reload repositories as they change, watchdog must be installed")
+    print("see https://pypi.python.org/pypi/watchdog/")
+    sys.exit()
+
+
+logger = logging.getLogger(__file__)
+
+def is_subdirectory(haystack, needle):
+    """
+    checks whether `needle` exists as a subdirectory of `haystack`
+    """
+    def func(arg, dirname, fnames):
+        if dirname == needle or needle in fnames:
+            logger.debug('%s is within %s', needle, haystack)
+            raise StopIteration()
+    try:
+        os.path.walk(haystack, func, None)
+    except StopIteration:
+        return True
+    return False
+
+
+class KlausEventHandler(FileSystemEventHandler):
+    """
+    Basic file-system event handler for a klaus app.
+    When git repositories are created, moved or deleted, watchdog will notify us here.
+    We try and update the klaus wsgi app with the modifications as they happen.
+    """
+    def __init__(self, klausapp, *args, **kwargs):
+        self.app = klausapp
+        return super(KlausEventHandler, self).__init__(*args, **kwargs)
+
+    def _add_dir(self, directory):
+        """
+        Attempts to register `directory` with the klaus app as a git repository.
+        If `directory` is not a git repo, this method does nothing.
+        """
+        try:
+            repo = FancyRepo(directory)
+        except dulwich.errors.NotGitRepository:
+            logger.debug("Dulwich says '%s' is not a git repository", directory)
+            return
+
+        matching = filter(lambda existing_repo: existing_repo.name == repo.name, self.app.repos)
+        if matching:
+             logger.warning("skipping repo with duplicate name:%s", repo)
+             return
+        logger.info("Adding repository, '%s' at '%s'", repo.name, directory)
+        self.app.repos.append(repo)
+
+    def on_created(self, event):
+        # We're only interested in directories; A git repo is always a directory
+        if type(event) != DirCreatedEvent:
+            return
+        logger.debug("Received directory creation event for '%s'", event.src_path)
+        self._add_dir(event.src_path)
+        self.app.update_repos_list()
+
+    def on_deleted(self, event):
+        if type(event) != DirDeletedEvent:
+            return
+        matching = filter(lambda existing_repo : existing_repo.path == event.src_path, self.app.repos)
+
+        map(lambda x: self.app.repos.remove(x), matching)
+        logger.info("%s was deleted; removing from klaus", event.src_path)
+        self.app.update_repos_list()
+
+    def on_moved(self, event):
+        """
+        Moved is not the same as deleted, unless the moved dir is no longer a subset of
+        any of the watched trees. In that case we need to consider it a deletion,
+        and remove it from the list of watched repos; otherwise we need to update its path.
+        This is a potentially expensive operation, because the only reliable way
+        to detect whether it is a subset (including symlinks etc) is to walk the path.
+        In practice, this shouldn't affect performance in the general case,
+        as moving a repository is a rare event.
+        """
+
+        matching = filter(lambda existing_repo : existing_repo.path == event.src_path, self.app.repos)
+        if not matching:
+            logger.debug("ignored non-repo move, %s -> %s", event.src_path, event.dest_path)
+            return
+
+        #first unregister old repositories
+        map(lambda x: self.app.repos.remove(x), matching)
+        # then see if the new destination is part of the klaus watchlist
+        for directory in self.app.basedirs:
+            if is_subdirectory(directory, event.dest_path):
+                logger.info("'%s' -> '%s'; updating references", event.src_path, event.dest_path)
+                self._add_dir(event.dest_path)
+
+        self.app.update_repos_list()
+        logger.info(
+            "%s was moved out of KLAUS_REPOS to %s, and is no longer a watched repository",
+            event.src_path,
+            event.dest_path
+        )
+
+
+class KlausAutoDiscover(Klaus):
+    """
+    Instead of the normal Klaus behaviour of receiving a list of git repositories
+    as `os.environ['KLAUS_REPOS']`, this app searches `os.environ['KLAUS_REPOS']`
+    for valid git repositories and updates the application automatically when any 
+    changes are made.
+    This simplifies administration significantly, but may cause problems with
+    large repositories.
+    """
+    def __init__(self, basedirs, site_name, use_smarthttp, **kwargs):
+        self.basedirs = basedirs
+        repo_paths = []
+        # first find all the existing repos in the given base directory
+        for directory in basedirs:
+            for fspath in os.listdir(directory):
+                try:
+                    path = os.path.join(directory, fspath)
+                    FancyRepo(path)  # The result is discarded. This is just a test.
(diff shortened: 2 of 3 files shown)


jonashaag:
Thanks for the patch. Have you tried this code? https://github.com/jonashaag/klaus/wiki/Autoreloader

ldrumm:
I have tried the existing Autoreloader.  However, it doesn't do intelligent autodiscovery of git repos, and if a directory is added that is not a git repository, the behaviour is likely to be a crash, or at least a stern warning with an HTTP500 from the wsgi server.  Additionally, the polling implementation, while perfectly efficient, doesn't  update klaus's list of repos immediately.  I think they both have valid use cases, but I believe the features introduced by his PR make administration a lot simpler.  

That said, if there are any improvements you wish me to make before you're willing to merge, let me know, and I''ll get right on it.

jonashaag:
OK, I've taken the general idea into consideration and I think it's a nice addition.

I hope you have the patience to do a few iterations on coding style, as this patch will need quite a bit of work to meet my standards :sunglasses:

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,165 @@
...
+        Attempts to register `directory` with the klaus app as a git repository.
+        If `directory` is not a git repo, this method does nothing.
+        """
+        try:
+            repo = FancyRepo(directory)
+        except dulwich.errors.NotGitRepository:
+            logger.debug("Dulwich says '%s' is not a git repository", directory)
+            return
+
+        matching = filter(lambda existing_repo: existing_repo.name == repo.name, self.app.repos)
```
jonashaag: This (and all the other places as well, of course) should use `.repo_map` to avoid all the cumbersome abuses of the list datastructure :-)

Please get rid of the `.repos` attribute in `Klaus` entirely. I don't know why it's there (probably for historic reasons) -- it's not needed at all. Perhaps it served as a way to keep repo order in the past, but that use case doesn't exist anymore because the repo list is ordered explicitly in the one view it's used. We can simply drop it in favor of the `repo_map` attribute. Please make this a separate commit as it is not really related to the autodiscovery thing.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,165 @@
...
+        if matching:
+             logger.warning("skipping repo with duplicate name:%s", repo)
+             return
+        logger.info("Adding repository, '%s' at '%s'", repo.name, directory)
+        self.app.repos.append(repo)
+
+    def on_created(self, event):
+        # We're only interested in directories; A git repo is always a directory
+        if type(event) != DirCreatedEvent:
```
jonashaag: This may be expressed as `event.is_directory`

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,165 @@
...
+    def on_moved(self, event):
+        """
+        Moved is not the same as deleted, unless the moved dir is no longer a subset of
+        any of the watched trees. In that case we need to consider it a deletion,
+        and remove it from the list of watched repos; otherwise we need to update its path.
+        This is a potentially expensive operation, because the only reliable way
+        to detect whether it is a subset (including symlinks etc) is to walk the path.
+        In practice, this shouldn't affect performance in the general case,
+        as moving a repository is a rare event.
+        """
```
jonashaag: Couldn't the symlink problem be solved by simply taking the `os.path.realpath` of the destination path and testing if that real path is a subpath of one of the basedirs? (by string comparison)

klaus/contrib/autodiscover.py
ldrumm: That's a lot better.  I didn't see that in the `watchdog` API.  Agreed.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,165 @@
...
+                try:
+                    path = os.path.join(directory, fspath)
+                    FancyRepo(path)  # The result is discarded. This is just a test.
+                    repo_paths.append(path)
+                except dulwich.errors.NotGitRepository:
+                    continue
+        self.observer = Observer()
+        fs_event_handler = KlausEventHandler(self)
+        # Then setup up a handler to notify us when the content of any of those directories change.
+        map(lambda directory: self.observer.schedule(fs_event_handler, directory, recursive=True), basedirs)
```
jonashaag: Please don't use `map` for functions that have side effects.

Also, if I understand this correctly, the watching goes recursively without a maximum depth level. This is going to be a problem as repositories may have very deeply nested directory hierarchies. Could the recursion level be limited somehow? Maybe we don't even need recursion. Maybe we're fine with only looking at the first level? I understand people may have complex repository hierarchies. But klaus doesn't need to support every extreme border case.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    def __init__(self, basedirs, site_name, use_smarthttp, **kwargs):
+        self.basedirs = basedirs
+        repo_paths = []
+        # first find all the existing repos in the given base directory
+        for directory in basedirs:
+            for fspath in os.listdir(directory):
+                try:
+                    path = os.path.join(directory, fspath)
+                    FancyRepo(path)  # The result is discarded. This is just a test.
```
jonashaag: This should be unified with the code from line 54

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self.app.update_repos_list()
+        logger.info(
+            "%s was moved out of KLAUS_REPOS to %s, and is no longer a watched repository",
+            event.src_path,
+            event.dest_path
+        )
+
+
+class KlausAutoDiscover(Klaus):
```
jonashaag: This whole initialization business doesn't even need a subclass since none of the internal logic of the `Klaus` class is changed. I think it's more elegant to just use the normal `Klaus` instance and do the watchdog stuff in a bootstrapping function.

klaus/contrib/autodiscover.py
ldrumm: This makes a lot of sense.  I must confess, I didn't actually fully investigate how `.repos` is used internally as this patch was initially just a proof of concept.  However a quick `grep -R -E '\.repos'` shows:

``` text
__init__.py:        self.repos = map(FancyRepo, repo_paths)
__init__.py:        self.repo_map = dict((repo.name, repo) for repo in self.repos)
__init__.py:            dict(('/'+repo.name, repo) for repo in app.repos)
views.py:    repos = sorted(current_app.repos, key=sort_key, reverse=reverse)
```
While we're dealing with `klaus.Klaus.repo_map`, i'd like to propose to change the dictionary key from `FancyRepo.name` to `FancyRepo.path` as `FancyRepo.name` does not guarantee correct behaviour for multiple repositories with the same basename.

Here we can key `.repo_map[repo.path]` to check for membership, and do a simple `del klausapp.repo_map[repo.path]` to remove an item.

I'm quite happy to remove the `.repos` attribute from the remaining parts of *klaus* that use it in another commit.
Interested to learn what you think about keying off `FancyRepo.path` instead of `FancyRepo.name`

klaus/contrib/autodiscover.py
ldrumm: I understand that a bootstrapping function might work just fine, but since we're adding custom attributes, `.observer`, `.basedirs`, and a custom `__del__` method for cleanup, we are in fact altering the behaviour of the `klaus.Klaus` enough to justify sub-classing.  If we were using a bootstrapping function and not adding custom attributes, we'd have to store global state, somewhere else, which feels a lot less elegant

klaus/contrib/autodiscover.py
ldrumm: Definitely

klaus/contrib/autodiscover.py
ldrumm: I'll look into this.  From memory, I think I encountered dulwich not detecting repositories if recursion was turned of completely, but I'll have to double check.

re `map`.  Yes.  I was just being lazy.  I'll refactor that.

klaus/contrib/autodiscover.py
jonashaag: That makes a lot of sense. Didn't think of that. Agreed!

klaus/contrib/autodiscover.py
jonashaag: Correct me if I'm wrong: `observer` and `basedirs` aren't going to change, right? `basedirs` is just used for constructing `observer`, which in turn is not used at all besides for watchdog setup (which is a one-time operation).

Cleanup is a good point. I have no solution other than `__del__` at this point. I'll give it a thought tomorrow. (`__del__` isn't guaranteed to be called, btw. The chances are good but no guarantees. But I don't see an alternative right now.)

klaus/contrib/autodiscover.py
jonashaag: Making this configurable and default to `False` seems reasonable.

klaus/contrib/autodiscover.py
ldrumm: Working on this now...

klaus/__init__.py
```diff
@@ -15,9 +15,8 @@ class Klaus(flask.Flask):
         'undefined': jinja2.StrictUndefined
     }
 
-    def __init__(self, repo_paths, site_name, use_smarthttp):
-        self.repos = map(FancyRepo, repo_paths)
-        self.repo_map = dict((repo.name, repo) for repo in self.repos)
+    def __init__(self, repo_paths, site_name, use_smarthttp, **kwargs):
+        self.repo_map = dict((repo.path, repo) for repo in map(FancyRepo, repo_paths))
```
jonashaag: `dict((path, FancyRepo(repo)) for path in repo_paths))` :-)

ldrumm: derp

klaus/contrib/autodiscover.py
ldrumm: FYI, I did some more thinking about this, and tried to code around the fact the `repo_map.keys()` are used in the URL to uniquely identify a repository.  I've left it for now.  If you can think of a way to uniquely identify repositories with the same basename that doesn't leak filesystem information in the URL, and keeps the URLs as pretty as they are now, I think this could be revisited, but for now, we'll just have to live with only using repositories that have unique basenames.

jonashaag: True. Still I think `path` is the correct key into that dictionary

ldrumm:
@jonashaag Just to let you know, I've not forgotten about this (not in the slightest), and will soon have time to improve this PR to the point that we agree it should be merged. Some good news is that I've been running a variation on this patch in production for a few months, with no issues, so I remain confident that this is a good long term option.

Please assign me if you're willing and able.

ldrumm:
Hi @jonashaag I've made some more cleanups, and merged your master branch. I think things look quite different to my initial PR and I believe they're ready for merge, so if you get some time, I'd welcome your comments.
If things are looking a bit messy, I'm happy to close this PR, and open new one with a single squashed commit.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+
+def is_subdirectory(haystack, needle):
+    """
+    checks whether `needle` exists as a subdirectory of `haystack`
+    """
+    haystack = os.path.realpath(haystack)
+    needle = os.path.realpath(needle)
+    if not os.path.isdir(needle):
+        return False
+    return needle.startswith(haystack)
```
jonashaag: You'll have to make sure both `haystack` and `needle` end with a `/` for the `.startswith` to work correctly. Counter example with the current implementation: `is_subdirectory("/foo", "/foo2")` is `True` if both paths exists, but should be False

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        if not event.is_directory:
+            return
+        self._klausapp.remove_repo(event.src_path)
+        self._klausapp.add_repo(event.src_path)
+
+    def on_created(self, event):
+        # We're only interested in directories; A git repo is always a directory
+        if not event.is_directory:
+            return
+        logger.debug("Received directory creation event for '%s'", event.src_path)
```
jonashaag: Use `%r`

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        self._klausapp.add_repo(event.src_path)
+
+    def on_deleted(self, event):
+        if not event.is_directory:
```
jonashaag: Copy comment from method above

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        a deletion, and remove it from the list of watched repos; otherwise we
+        need to update its path. This is a potentially expensive operation,
+        but shouldn't affect performance in the general case, as moving a
+        repository is a rare event.
+        """
+        logger.debug(
+            "Received repository move event:'%s'->'%s'",
+            event.src_path,
+            event.dest_path
+        )
```
jonashaag: Try to keep the visual space occupied by logging lines as small as possible. Condense this to a single line. It's fine if it's 100 chars wide or so.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        # then see if the new destination is part of the klaus watchlist
+        for directory in self._klausapp.basedirs:
+            if is_subdirectory(directory, event.dest_path):
+                logger.info(
+                    "'%s' -> '%s'; updating references",
+                    event.src_path,
+                    event.dest_path
+                )
+                self._klausapp.add_repo(event.dest_path)
+                break
```
jonashaag: Use `return` here

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+                event.src_path,
+                event.dest_path
+            )
+
+
+class KlausAutoDiscover(Klaus):
+    """
+    Instead of the normal Klaus behaviour of receiving a list of git
+    repositories as `os.environ['KLAUS_REPOS']`, this searches
+    `os.environ['KLAUS_REPOS']` for valid git repositories and updates klaus
```
jonashaag: Let's rename this to `KLAUS_REPOS_ROOTS`, similar to the `KLAUS_REPOS_ROOT` in the autoreloader

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+    automatically when any changes are made. if `recursive=True`, klaus will
```
jonashaag: This doesn't explain that repos are also looked for in each directory in `KLAUS_REPOS_ROOTS`, i.e. that `KLAUS_REPOS_ROOTS` is a list of directories containing repositories.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+    watch every subdirectory in the given directories, and discover all git
+    repositories therein. This simplifies administration significantly, but may
+    cause problems with large repositories, as the underlying file watch
+    mechanisms don't expose a recursion depth parameter on any of the common
+    platforms (Linux, win32, OS X).
+    """
+    def __init__(self, basedirs, site_name, use_smarthttp, **kwargs):
+        recursive = kwargs.pop('discover_recursive', False)
```
jonashaag: Name this variable `discover_recursive`

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        super(KlausAutoDiscover, self).__init__([], site_name, use_smarthttp, **kwargs)
+        self._lock = threading.RLock()
```
jonashaag: Explain what the lock is being used for.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        self._basedirs = basedirs
+        self._observer = Observer()
+
+        # First find and add all the existing repos in the given base directory.
```
jonashaag: "base directorIES"

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+    repositories therein. This simplifies administration significantly, but may
+    cause problems with large repositories, as the underlying file watch
+    mechanisms don't expose a recursion depth parameter on any of the common
+    platforms (Linux, win32, OS X).
+    """
+    def __init__(self, basedirs, site_name, use_smarthttp, **kwargs):
+        recursive = kwargs.pop('discover_recursive', False)
+        super(KlausAutoDiscover, self).__init__([], site_name, use_smarthttp, **kwargs)
+        self._lock = threading.RLock()
+        self._basedirs = basedirs
```
jonashaag: Rename `basedirs` to `root_dirs` to keep it consistent with `KLAUS_REPOS_ROOTS`

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        self._observer = Observer()
+
+        # First find and add all the existing repos in the given base directory.
+        for directory in basedirs:
+            for fspath in os.listdir(directory):
```
jonashaag: Rename loop vars:

```py
for root_dir in root_dirs:
    for maybe_repo in os.listdir(root_dir):
```

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+                path = os.path.join(directory, fspath)
+                self.add_repo(path)
+
+        # Then create a handler to notify us of any changes within those dirs.
+        fs_event_handler = KlausFSEventHandler(self)
+        for directory in basedirs:
```
jonashaag: Same here

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+        If `directory` is not a git repo, this method does nothing.
+        """
+        with self._lock:
+            try:
+                repo = FancyRepo(directory)
+            except dulwich.errors.NotGitRepository:
+                logger.debug("Dulwich says '%s' is not a git repository", directory)
+                return
+
+            if self.repo_map.get(repo.name):
```
jonashaag: `repo.name in self.repo_map`

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+                logger.debug("skipping repo with duplicate name:%s", repo)
+                return
+            logger.info("Adding repository, '%s' at '%s'", repo.name, directory)
+            self.repo_map[repo.name] = repo
+
+    def remove_repo(self, directory):
+        """
+        given a filesystem path, `directory` attempts to remove the
+        corresponding repository from the list of watched repositories.
+        """
```
jonashaag: This should be split into two separate methods: One that checks whether we have the directory in our list, and one that assumes we have it in our list and removes it.

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+            key_map = dict(zip(
+                (repo.path for repo in self.repo_map.values()),
+                self.repo_map)
+            )
+            try:
+                name = key_map[directory]
+                del self.repo_map[name]
+                return True
+            except KeyError:
+                logger.debug(
```
jonashaag: This can be simplified to an `if/else` by using `directory in key_map`

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+    def remove_repo(self, directory):
+        """
+        given a filesystem path, `directory` attempts to remove the
+        corresponding repository from the list of watched repositories.
+        """
+        with self._lock:
+            key_map = dict(zip(
+                (repo.path for repo in self.repo_map.values()),
+                self.repo_map)
+            )
```
jonashaag:
```py
def get_repo_by_path(self, path):
    for repo in self.repo_map.values():
        if repo.path == path:
            return repo
```

Then `remove_repo` can be implemented as

```py
def remove_repo(self, repo):
    del self.repos_map[repo.name]
```

or

```py
def remove_repo(self, directory):
    del self.repos_map[self.get_repo_by_path(directory).name]
```

klaus/contrib/autodiscover.py
```diff
@@ -0,0 +1,194 @@
...
+                )
+                pass
+
+    def __del__(self):
+        self._observer.stop()
+        self._observer.join()
+
+
+def _app(htdigest_file=None):
+    return make_app(
```
jonashaag: Putting the code common to the both cases into a function is a good idea, but the pattern used elsewhere (`wsgi.py` and `wsgi_autoreload.py`) is to repeat it. Can we do this particular cleanup after in a different PR?

klaus/views.py
```diff
@@ -21,7 +21,7 @@ def repo_list():
     else:
         sort_key = lambda repo: repo.name
         reverse = False
-    repos = sorted(current_app.repos, key=sort_key, reverse=reverse)
+    repos = sorted(current_app.repo_map.values(), key=sort_key, reverse=reverse)
```
jonashaag: Also this cleanup: different PR

ldrumm:
closing in favour of #267

PR #117, opened 2015-04-28
Fix errant .split() calls that causes a TypeError

### LINKED ISSUE #116: Bug in contrib/wsgi_autoreload.py
#91 introduced a bug in the wsgi_autoreload component which leads to the error "TypeError: coercing to Unicode: need string or buffer, list found".

Steps to reproduce:
 - Follow the setup steps at https://github.com/jonashaag/klaus#using-a-real-server, but use klaus.contrib.wsgi_autoreload, and specify one path for KLAUS_REPOS

Removing the .split() from L55 and L62 of this file seems to fix this

insertjokehere:
Fix errant .split() calls that causes a TypeError when trying to use the wsgi_autoreload component

Fixed #116


jonashaag:
Thanks for the patch!

PR #124, opened 2015-06-24
allow to specify repositories root

yedpodtrzitko:
hi,
it looks like `make_autoreloading_app` doesnt really work right now - it accepts `KLAUS_REPOS` as a param (which is supposed to be a list of directories), but when it spawns a thread to monitor changes later, it expects just a single directory there.

So I've added another env variable `KLAUS_REPOS_ROOT`, which can be used for specifying a parent directory for repositories in it, instead of defining a list of repositories explicitly.

This change is backward compatible, as it fallbacks to `KLAUS_REPOS` in case `KLAUS_REPOS_ROOT` is not defined.

### DIFF
--- a/klaus/contrib/wsgi_autoreload.py
+++ b/klaus/contrib/wsgi_autoreload.py
@@ -1,8 +1,9 @@
 import os
 import time
 import threading
-from klaus import make_app
+import warnings
 
+from klaus import make_app
 
 # Shared state between poller and application wrapper
 class _:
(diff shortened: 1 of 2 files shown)


jonashaag:
Thanks a lot for the patch!

I agree that the environment variable should have a different name for the autoreloader to make it very clear that a single root directory is expected (not a list of directories).

The code in `klaus.contrib.*` should be self-contained and as simple as possible as it's supposed to serve as a basis for customization/"copy-&-paste programming". Please don't extract code from there to the main code base (I'm talking about `klaus.utils.get_repos`).

Instead, please add a `DeprecationWarning` if `KLAUS_REPOS` is used in `wsgi_autoreload.py`.

yedpodtrzitko:
@jonashaag something like this?

jonashaag:
That's better.

- The example in the README isn't for the autoreloader but for the "list of repositories" case
- The warning shouldn't say "for WSGI apps" but "for the autoreloader"
- `.get('KLAUS_REPOS')` should be `['KLAUS_REPOS']` (should fail with `KeyError` if both aren't present)

yedpodtrzitko:
> The warning shouldn't say "for WSGI apps" but "for the autoreloader"

done

> .get('KLAUS_REPOS') should be ['KLAUS_REPOS'] (should fail with KeyError if both aren't present)

done

> The example in the README isn't for the autoreloader but for the "list of repositories" case. Don't change that. Update the wiki page accordingly: https://github.com/jonashaag/klaus/wiki/Autoreloader

done

klaus/contrib/wsgi_autoreload.py
```diff
@@ -49,18 +50,20 @@ def app(environ, start_response):
...
 
 
+if 'KLAUS_REPOS' in os.environ:
+    warnings.warn("use KLAUS_REPOS_ROOT instead of KLAUS_REPOS for the autoreloader apps", DeprecationWarning)
+
 if 'KLAUS_HTDIGEST_FILE' in os.environ:
     with open(os.environ['KLAUS_HTDIGEST_FILE']) as file:
-        application = make_app(
-            os.environ['KLAUS_REPOS'],
+        application = make_autoreloading_app(
```
jonashaag: Thanks for fixing this, this was an oversight from 665a7292ab68fecbc337d116a3d91598bf7714de

jonashaag:
LGTM! Thanks!

PR #125, opened 2015-06-25
Tarutils Python 2.6 compatibility

danajpl:
tarutils.py breaks under python 2.6. This workaround fixes it.


jonashaag:
Thanks! I also added a test for that bug in 87993de134649620fe63dc01aa6573086d982fc9

danajpl:
Sweet!

PR #128, opened 2015-06-26
fix autoreload condition

yedpodtrzitko:
this assignment is wrong - it's `old_contents` which needs to be reassigned, not `new_contents`


jonashaag:
OMG. Thanks for the patch.

Issue #126, opened 2015-06-25
Show committer in revision overview


jelmer:
It would be nice if klaus showed the committer in the revision overview, if it is different from the author.

(I hope this is an actual issue, and not another dupe...)

jonashaag:
Accepted. Patches welcome.

Issue #122, opened 2015-06-24
Display clone URL


jelmer:
It would be great if klaus could display clone instructions for each repository, if known.

If the smart server is enabled, it could display the URL for that. In other situations, it could retrieve one from the repository config.

jonashaag:
This is already implemented for Smart HTTP. Where exactly do you get the URL from in the other case?

jelmer:
On Thu, Jun 25, 2015 at 12:49:36AM -0700, Jonas Haag wrote:
> This is already implemented for Smart HTTP. Where exactly do you get the URL from in the other case?
I was thinking either a setting for the public URL in the git config file, or
alternatively some sort of translation setting in the klaus config. I.e.:

PUBLIC_URL = lambda r: r.replace("/srv/git/", "git://jelmer.uk/")

Jelmer

jonashaag:
I think we could support `.git/cloneurl` file, `gitweb.url` and `@git_base_url_list` options. The relationship of these options has to be explained to me in simple English as I couldn't figure it out.

PR #131, opened 2015-07-19
Cope with alternative name for flask-httpauth on Debian.

jelmer:


jonashaag:
Um, this isn't the `flask-httpauth` package we're using here, but https://github.com/jonashaag/httpauth

jelmer:
Argh. Clearly all my klaus issue reports are doomed. 

Sorry.

jonashaag:
This must be horrible feeling… if it is any consolation, #126 and #122 are both valid and useful suggestions. :-D

Issue #67, opened 2013-07-01
git blame equivalent


jonashaag:

jelmer:
FWIW https://github.com/jelmer/dulwich/issues/245 tracks blame support in dulwich.

jonashaag:
Thanks, subscribed.

jonashaag:
I'm fine with the implementation above for now (no real plans for switching to libgit2), so if anyone's interested in resurrecting that patch, you're welcome to work on it :-)

PR #132, opened 2015-07-27
Feature/git blame

posativ:
#67, patch set seems intact. Feel free to use it.


jonashaag:
Cool!

TODO @jonashaag:

- [ ] Refactor context (`is_binary`, `is_image` code)
- [ ] Add tests that make sure we render binary/image/too large correctly for both commit and blame

PR #134, opened 2015-09-10
ctags support

jonashaag:
Experimental new feature. Looking for feedback! @posativ @jelmer @lb1a @danajpl @ewdurbin @yedpodtrzitko

EDIT: You'll have to install ctags and python-ctags

TODO:

- [x] In multi-language code bases, prefer tag matches of the language of the current document, as they're more likely to be the right candidate.
- [x] Handle case where running ctags takes a long time (background? git hook? prefill cache?)
  POSTPONED #135
- [x] Integrate into CLI
- [x] Fix CSS
- [x] Handle python-ctags not installed case in CLI
- [x] Handle ctags not installed case in CLI
- [x] Restrict revisions to be ctagged by default as to not open a DDOS security issue
- [x] Ensure ``ctags`` binary found is EXUBERANT ctags
- [x] Add tests
- [x] Wiki entry
- [x] Split into multiple commits (refactoring vs. feature)


jelmer:
Whoa, that's pretty nice! :) And it is surprising how few lines of code you need.

Some other possible TODOs:

 * If ctags isn't installed, subprocess.check_call() raises OSError, which causes a nondescript 500 page and no errors in the logs
 * I'd want the compressed cache on disk rather than in memory; I don't think having both a compressed and non-compressed variant in memory is worth the complexity. gzip isn't that expensive.
 * it would be nice to have a way to disable it completely, for some branches or for e.g. non-tip revisions. Requesting all revisions in the Samba repository on my server seems like a good way to DDOS it otherwise :)
 * optionally, a script that can run as a post-receive hook and that updates the disk cache (so that this doesn't have to be done online).

jonashaag:
> I'd want the compressed cache on disk rather than in memory; I don't think having both a compressed and non-compressed variant in memory is worth the complexity. gzip isn't that expensive.

Both caches are on disk because python-ctags doesn't support reading from memory. I've added a comment about that.

> it would be nice to have a way to disable it completely, for some branches or for e.g. non-tip revisions. Requesting all revisions in the Samba repository on my server seems like a good way to DDOS it otherwise :)

Good point. I'd like to keep the amount of command line options related to ctags to a minimum. And I'd like to keep the implementation complexity to a minimum. So maybe we can find a smart default setting for what revisions we're ctagging? How about we default to "all branches" and have an option to enable it for all revisions?

> optionally, a script that can run as a post-receive hook and that updates the disk cache (so that this doesn't have to be done online).

Hm. If we let klaus manage the hook, the only improvement is that is saves us a few seconds of response time if the tagsfile isn't there yet. And I don't like users to deal with this stuff, because of my "It Just Works" ambitions :)

jelmer:
On Fri, Sep 11, 2015 at 02:01:59AM -0700, Jonas Haag wrote:
> > I'd want the compressed cache on disk rather than in memory; I don't think having both a compressed and non-compressed variant in memory is worth the complexity. gzip isn't that expensive.
> 
> Both caches are on disk because python-ctags doesn't support reading from memory. I've added a comment about that.
AFAICT it uses lru cache from Dulwich for both, which is in memory.

> > it would be nice to have a way to disable it completely, for some branches or for e.g. non-tip revisions. Requesting all revisions in the Samba repository on my server seems like a good way to DDOS it otherwise :)
> 
> Good point. I'd like to keep the amount of command line options related to ctags to a minimum. And I'd like to keep the implementation complexity to a minimum. So maybe we can find a smart default setting for what revisions we're ctagging? How about we default to "all branches" and have an option to enable it for all revisions?
Defaulting to all branches seems readonable to me. 

> > optionally, a script that can run as a post-receive hook and that updates the disk cache (so that this doesn't have to be done online).
> 
> Hm. If we let klaus manage the hook, the only improvement is that is saves us a few seconds of response time if the tagsfile isn't there yet. And I don't like users to deal with this stuff, because of my "It Just Works" ambitions :)
I do sympathise with that, but it's only going to work for smaller repositories. For Samba, cloning and running ctags takes > 10 seconds, making it impractical to run on the fly.

Jelmer

jonashaag:
> AFAICT it uses lru cache from Dulwich for both, which is in memory.

The entries that are put into the Dulwich LRU cache are only filenames of tagsfiles in temporary locations on disk.

klaus/ctagscache.py
```diff
@@ -0,0 +1,185 @@
...
+    :return: path to the uncompressed version of the tagsfile
+    """
+    _, uncompressed_tagsfile_path = tempfile.mkstemp()
+    with gzip.open(compressed_tagsfile_path, 'rb') as compressed:
+        with open(uncompressed_tagsfile_path, 'wb') as uncompressed:
+            shutil.copyfileobj(compressed, uncompressed)
+    return uncompressed_tagsfile_path
+
+
+MiB = 1024 * 1024
```
yedpodtrzitko: this constant would looks much better on the beginning of the file

PR #137, opened 2015-10-10
prefix Raw diff link with SITE_NAME

johnko:
Ref: #136

Issue #136, opened 2015-10-10
if serving from a suburi (eg. '/git'), raw diff url doesn't include the suburi


johnko:
Example at https://v.johnko.ca/git/rad/commit/ec65ebb7128dd7dd117481783228674a2726b119/

Click on "Raw diff" which goes to https://v.johnko.ca/rad/commit/ec65ebb7128dd7dd117481783228674a2726b119.diff

Note it's missing '/git'

Did I do something wrong?

johnko:
More info:

I'm using
```
make_app(
            [os.path.join(repos_root, x) for x in os.listdir(repos_root)],
            "v.johnko.ca/git",
            "1",
            None
        )
```

johnko:
I think it has to do with me proxying with NginX since the Raw Diff url  is made with

```
<a href="{{ request.path.rstrip('/') }}.diff">Raw diff</a>
```

I'll see what gets passed from NginX to the klaus socket

johnko:
Just as a side note:
When using NGINX, I don't pass `X-Script-Name` because when I do it seems to break repos which names start with the same path.

For example:
If I set this in NGINX `proxy_set_header X-Script-Name '/git';`
then repos named `github-backup` or `gitlab-ce` were stripped of 'git' and became `hub-backup` or `lab-ce`

jonashaag:
See comment in #137 for further discussion of this topic.

As for the `SCRIPT_NAME` thing, if you're using uWSGI, see http://flask.pocoo.org/docs/0.10/deploying/uwsgi/. Otherwise I'd be curious to know what your setup is.

jonashaag:
Fixed in https://github.com/jonashaag/klaus/commit/0a286f5110665f6a3781d52905527bf9a4c2877f and released as 0.7.1

PR #138, opened 2015-10-10
fix Raw diff link

johnko:
second attempt
Ref: #136 , #137


jonashaag:
What happens if you use `'patch'` as first argument to `url_for` instead of appending `.diff` manually? If that works that would be the recommended way to do it.

johnko:
Good call

PR #141, opened 2015-11-01
Disable GPG signing during tests.

jelmer:


jonashaag:
What use is this?

jelmer:
On Sun, Nov 01, 2015 at 12:02:19PM -0800, Jonas Haag wrote:
> What use is this?
I've got GPG signing enabled by default in my Git configuration, which makes
running the Klaus testsuite painful as I have to manuall approve each GPG
signature.

Jelmer

jonashaag:
I think we should rather set `commit.gpgSign` in the repo's config

jelmer:
On Sun, Nov 01, 2015 at 02:41:41PM -0800, Jonas Haag wrote:
> I think we should rather set `commit.gpgSign` in the repo's config
Ideally, just set a custom user config so that configuration from
$HOME/.gitconfig doesn't leak through in the tests.

jonashaag:
Right!

PR #149, opened 2015-12-09
Add klaus.1 manpage.

jelmer:
Not sure if you care about manpages, but if you do - here is one :)

### DIFF
--- a/klaus.1
+++ b/klaus.1
@@ -1,0 +1,55 @@
+.TH KLAUS "1" "December 2015" "klaus 4e82832" "User Commands"
+.SH NAME
+klaus \- easy to set up Git web viewer
+.SH SYNOPSIS
+.B klaus
+[\fIOPTION\fR]... [\fIDIR\fR]...
+.SH DESCRIPTION
+Klaus is a simple and easy-to-set-up Git web viewer that Just Works\(tm.
+.PP
+Note that the klaus binary just starts a test instance. The klaus script uses
+wsgiref internally which doesn't scale at all - it's single-threaded and
+non-asynchronous.
+.PP
+It supports syntax highlighting and Git Smart HTTP.
+.SH OPTIONS
+.TP
+\fB\-h\fR, \fB\-\-help\fR
+show this help message and exit
+.TP
+\fB\-\-host\fR HOST
+default: 127.0.0.1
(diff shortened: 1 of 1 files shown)


jonashaag:
Cool! Thanks for the patch! I do care for man pages, but I'm not sure how well they integrate with the Python packaging system? Are they made accessible from `man klaus` if installed in `/usr/lib/pythonX.Y/...` (or similar)? They're most certainly not accessible if installed in a virtualenv?

Btw, I think we should be using something like https://github.com/oz123/man-utils to avoid documentation duplication. I never used *man-utils* but it looks promising.

jelmer:
On Wed, Dec 09, 2015 at 12:58:39AM -0800, Jonas Haag wrote:
> Cool! Thanks for the patch! I do care for man pages, but I'm not sure how well they integrate with the Python packaging system? Are they made accessible from `man klaus` if installed in `/usr/lib/pythonX.Y/...` (or similar)? They're most certainly not accessible if installed in a virtualenv?
> 
> Btw, I think we should be using something like https://github.com/oz123/man-utils to avoid documentation duplication. I never used *man-utils* but it looks promising.
If you'd like to avoid shipping a troff file, what about just using
docutils' rst2man ? See
http://docutils.sourceforge.net/sandbox/manpage-writer/rst2man.txt

I'm adding the manpage for Debian, and man-utils doesn't appear to be
packaged in Debian yet.

Jelmer

jonashaag:
> I'm adding the manpage for Debian, and man-utils doesn't appear to be
> packaged in Debian yet.

Hm. It's not about shipping a troff file. It's about not repeating yourself w.r.t. the CLI options, and potentially missing updates if a command-line argument is changed. For the time being, it should be ok to add a note to `RELEASE_PROCESS`.

jonashaag:
There seems to be a missing closing parenthesis in the description: `\(tm` or does this have some other special meaning?

jelmer:
On Thu, Dec 10, 2015 at 01:49:19AM -0800, Jonas Haag wrote:
> There seems to be a missing closing parenthesis in the description: `\(tm` or does this have some other special meaning?
Yes, that's intentional - troff formats this a small supercase "TM".

Jelmer

PR #150, opened 2016-01-30
use modern google webfonts; improved font size

tschaefer:
Hi,

Great piece of software. I was not happy with the look and feel, so I replaced the fonts _sans-serif_ and _monospace_ with Google fonts _Ubuntu_ and _Inconsolata_ and adjusted some size settings. Maybe you like it and merge my pull request.

Cheers Tobias.

### DIFF
--- a/klaus/static/klaus.css
+++ b/klaus/static/klaus.css
@@ -3,12 +3,14 @@
 body, header, #content { overflow: auto; }
 
 /* Reset */
-body { margin: 0; padding: 0; font-family: sans-serif; }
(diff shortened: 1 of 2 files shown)


jonashaag:
Hi, thanks for the patch!

Nice fonts, typography and improved styling/CSS is something I wanted to do for a long time but unfortunately the current state is good enough for me, so no real motivation from my side :/

However I don't think we should be using any external Web fonts (neither include them in the distribution), I think it just doesn't make that much of a difference. Maybe we could do some font tweaking anyways by specifying something more than just `sans-serif`. (Steal GitHub's `font-family` set up maybe?)

PR #160, opened 2016-04-27
Python 3 support for ctags highlighting, fixes #159

### LINKED ISSUE #159: Python 3 compatibility
Hi, it seems that the Web UI works well with Python 3 and dulwich 0.13.0. However, Smart HTTP does not:
```
127.0.0.1 - - [26/Apr/2016 14:41:48] "GET /klaus/info/refs?service=git-upload-pack HTTP/1.1" 403 -
Error on request:
  File "/usr/lib64/python3.5/site-packages/werkzeug/serving.py", line 193, in run_wsgi
    execute(self.server.app)
  File "/usr/lib64/python3.5/site-packages/werkzeug/serving.py", line 184, in execute
    write(data)
  File "/usr/lib64/python3.5/site-packages/werkzeug/serving.py", line 164, in write
    assert isinstance(data, bytes), 'applications must write bytes'
AssertionError: applications must write bytes
```
I don't know if this is an issue with dulwich or klaus but I thought it would be nice to have an open bug to track the state of Python 3 compatibility.

sbraz:
I'll make a pull request to make it easier for you to see force-pushs.

### DIFF
--- a/klaus/highlighting.py
+++ b/klaus/highlighting.py
@@ -1,3 +1,5 @@
+from six.moves import filter
+
 from pygments import highlight
 from pygments.lexers import get_lexer_by_name, get_lexer_for_filename, \
                             guess_lexer, ClassNotFound, TextLexer
@@ -34,17 +36,18 @@ class KlausDefaultFormatter(HtmlFormatter):
 
     def _lookup_ctag(self, token):
         matches = list(self._get_all_ctags_matches(token))
-        best_matches = self.get_best_ctags_matches(matches)
+        best_matches = list(self.get_best_ctags_matches(matches))
         if not best_matches:
             return None, None
         else:
-            return best_matches[0]['file'], best_matches[0]['lineNumber']
+            return (best_matches[0]['file'].decode("utf-8"),
+                    best_matches[0]['lineNumber'])
 
     def _get_all_ctags_matches(self, token):
-        FIELDS = ('file', 'lineNumber', 'kind', 'language')
+        FIELDS = ('file', 'lineNumber', 'kind', b'language')
         from ctags import TagEntry
         entry = TagEntry()  # target "buffer" for ctags
-        if self._ctags.find(entry, token, 0):
+        if self._ctags.find(entry, token.encode("utf-8"), 0):
             yield dict((k, entry[k]) for k in FIELDS)
             while self._ctags.findNext(entry):
                 yield dict((k, entry[k]) for k in FIELDS)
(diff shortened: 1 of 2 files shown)


klaus/highlighting.py
```diff
@@ -34,17 +36,18 @@ def _format_lines(self, tokensource):
 
     def _lookup_ctag(self, token):
         matches = list(self._get_all_ctags_matches(token))
-        best_matches = self.get_best_ctags_matches(matches)
-        if not best_matches:
+        try:
+            best_match = next(self.get_best_ctags_matches(matches))
```
jonashaag: IMO this makes things more difficult to understand. I'd simply use a list comprehension in `get_best_ctags_matches `, so we're sure we always get a list here, and then use the `[0]` approach, which makes it very easy to see that we're only looking at the first element.

sbraz: Well using `next()` will be faster since it means we can stop before we have the complete list. To be honest, I'd have used it for `matches` too :)
But as it's your module, I'll use the list approach. I'll repush in a few minutes.

jonashaag: > Well using next() will be faster since it means we can stop before we have the complete list. To be honest, I'd have used it for matches too :)

I'd say this sounds like premature optimization, doesn't it? If you're interested in benchmarking the performance difference (could be a simple as trying to notice a difference in server response time), you're welcome; otherwise, I'd always propose to go with the unoptimized, simpler version of some piece of code.

> But as it's your module, I'll use the list approach. I'll repush in a few minutes.

I appreciate the thought but I think we should try to keep personal preference out of such discussions :-)

sbraz: I have not benchmarked it, I just noticed that you use `yield` so I assumed that it was to improve performance and not store everything in memory before returning from the function. Now that I think about it, using `yield` also makes the code nicer so maybe I just misinterpreted your choice.
I'm also happy with calling `list` if you think it is more clear. I use `next()` often so that may be why it does not seem unclear to me.
Anyway, I re-pushed with the list option, it probably makes little difference in the end :D

sbraz: I just looked at the length of the lists on an example page and the overhead is unnoticeable, so I guess I was overzealous indeed :)

jonashaag:
Awesome, thanks a lot!

sbraz:
@jonashaag Can you make a new release with the Python 3 support?

jonashaag:
Done!

PR #162, opened 2016-05-09
ctags: improve portability

sbraz:
I noticed while testing python-ctags that if a non-ASCII character is present in the file path, this would not work on Windows. I don't know how well python-ctags or klaus work on other platforms but this should be future-proof.


jonashaag:
:+1:

PR #168, opened 2016-08-23
Wsgi autoreload recursive

samdfonseca:
Ripped off `klaus/contrib/wsgi_autoreload.py` and modified it to find git repos contained in a directory recursively.

Useful when KLAUS_REPOS_ROOT has a layout such as:

```
|-- directory1
|    -- repo1
|-- directory2
|    -- repo2
|-- repo3
```

### DIFF
--- a/klaus/contrib/wsgi_autoreload_recursive.py
+++ b/klaus/contrib/wsgi_autoreload_recursive.py
@@ -1,0 +1,80 @@
+from __future__ import print_function
+import os
+import time
+import threading
+import warnings
+
+from klaus import make_app
+
+# Shared state between poller and application wrapper
+class _:
+    #: the real WSGI app
+    inner_app = None
+    should_reload = True
+
+
+def find_git_repos_recursive(dir):
+    repos = []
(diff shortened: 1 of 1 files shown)


jonashaag:
Hi Sam,

thanks for the contribution!

Why not simply flatten the nesting before invoking the reloader script? I.e., find the git folders recursively in a shell scripts and then invoke klaus with the paths found there?

s-ol:
This fails for bare repos (e.g. when stored as folder `repo.git`). I think even adding `or dirpath.endswith('.git')` is not robust enough of a check, and something that actually interacts with git would be better to use.

Since we know dulwich is available, we could use that for checking, or at least orient on their way of detecting the repository: https://github.com/dulwich/dulwich/blob/master/dulwich/repo.py#L1037, in pseudocode:

```
if isfile(join(current, ".git")) or isdir(join(current, ".git", "objects)):
  add_repo(current) # full rep
elif isdir(join(current, "objects")) and isdir(join(current, "refs")):
  add_repo(current) # bare repo
else:
  recurse()
```

PR #170, opened 2017-01-05
Cache htdigest file. Fixes #169

### LINKED ISSUE #169: Using htdigest file in wsgi_autoreload.py results in "I/O operation on closed file"
The htdigest file is opened like this in [wsgi_autoreload.py](https://github.com/jonashaag/klaus/blob/master/klaus/contrib/wsgi_autoreload.py)

    with open(os.environ['KLAUS_HTDIGEST_FILE']) as file:
        application = make_autoreloading_app(
            os.environ.get('KLAUS_REPOS_ROOT') or os.environ['KLAUS_REPOS'],
            os.environ['KLAUS_SITE_NAME'],
            os.environ.get('KLAUS_USE_SMARTHTTP'),
            file)

But `make_autoreloading_app` returns an inner function that does not call `klaus.make_app` until `application` is called. At that moment we are outside of the with statement and the file object has been closed:

```
def make_autoreloading_app(repos_root, *args, **kwargs):
    def app(environ, start_response):
        if _.should_reload:
            # Refresh inner application with new repo list
            print("Reloading repository list...")
            _.inner_app = make_app(
                [os.path.join(repos_root, x) for x in os.listdir(repos_root)],
                *args, **kwargs
            )
            _.should_reload = False
        return _.inner_app(environ, start_response)

    # Background thread that polls the directory for changes
    poller_thread = threading.Thread(target=(lambda: poll_for_changes(10, repos_root)))
    poller_thread.daemon = True
    poller_thread.start()

    return app
```

I could make a fix for this, and would also like to expose more kwargs as environment variables.

jimmyppi:
Also added some simple contrib tests.

### DIFF
--- a/klaus/contrib/wsgi_autoreload.py
+++ b/klaus/contrib/wsgi_autoreload.py
@@ -3,6 +3,7 @@ import os
 import time
 import threading
 import warnings
+from io import open, StringIO
 
 from klaus import make_app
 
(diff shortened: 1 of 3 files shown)


jonashaag:
Awesome! Thanks a lot!

PR #171, opened 2017-01-08
Expose all app arguments as env variables

jimmyppi:
TODO: How do you want to handle boolean arguments? Now everything becomes True if the env variable contains something other than the empty string.

### DIFF
--- a/tests/test_contrib.py
+++ b/tests/test_contrib.py
@@ -11,25 +11,23 @@ from .utils import *
 from .test_make_app import can_reach_unauth, can_push_auth
 
 
-def clear_env():
-    for var in list(os.environ):
-        if var.startswith('KLAUS_'):
-            os.environ.pop(var)
-
-
 def check_env(env, expected_args, expected_kwargs):
-    clear_env()
     os.environ.update(env)
     args, kwargs = app_args.get_args_from_env()
     assert args == expected_args
     assert kwargs == expected_kwargs
 
 
-def test_app_args_from_env():
-    clear_env()
+def test_missing_in_env(monkeypatch):
+    """Test that KeyError is raised when required env var is missing"""
+    monkeypatch.setattr(os, 'environ', os.environ.copy())
     with pytest.raises(KeyError):
         args, kwargs = app_args.get_args_from_env()
 
+
+def test_minimum_env(monkeypatch):
+    """Test to provide only required env var"""
+    monkeypatch.setattr(os, 'environ', os.environ.copy())
     check_env(
         {'KLAUS_SITE_NAME': TEST_SITE_NAME},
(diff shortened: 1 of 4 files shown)


jonashaag:
Thanks for the patch!

I don't get what you mean by the pull request title.

jimmyppi:
This patch adds support for these environment variables:

  KLAUS_REQUIRE_BROWSER_AUTH
  KLAUS_DISABLE_PUSH
  KLAUS_UNAUTHENTICATED_PUSH
  KLAUS_CTAGS_POLICY

I would like your opinion about how to handle the boolean arguments.
Should for example `KLAUS_REQUIRE_BROWSER_AUTH=0` be transformed to False?

jonashaag:
I see. Not sure about the booleans. How do other programs do this? I guess it's safe to say case-insensitive "yes", "true", "1" are True and "no", "false", "0" are false; however it's also fine to only support one of them as long as we reject unrecognised input.

jimmyppi:
This looks perfect:

https://docs.python.org/2/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool

```
    Convert a string representation of truth to true (1) or false (0).

    True values are y, yes, t, true, on and 1; false values are n, no, f, false, off and 0. Raises ValueError if val is anything else.
```

jonashaag:
Can we be sure distutils is available? Otherwise we could also simply copy and paste that function.

jimmyppi:
It is included in the standard library in both py2 and py3

jonashaag:
:+1:

jonashaag:
Cool thanks!

Would you mind using monkeypatching (see my comment) for the `os.environ` related tests?

jimmyppi:
Sure! What comment are you referring to?

jonashaag:
https://github.com/jonashaag/klaus/pull/171#pullrequestreview-15631331

jimmyppi:
Strange, I cannot see any review.

```
Reviewers

No reviews 
```

Maybe a github issue?

jonashaag:
A damn I forgot to submit

tests/test_contrib.py
```diff
@@ -4,34 +4,91 @@
...
 
+import pytest
+
+from klaus.contrib import app_args
 from .utils import *
 from .test_make_app import can_reach_unauth, can_push_auth
 
 
+def clear_env():
+    for var in list(os.environ):
```
jimmyppi: Had to use list() because os.environ becomes a generator in py3, which should not be modified while iterating over it.

But yes, I'll use monkeypatch instead, thanks.

jonashaag:
Awesome, thanks a lot!

Issue #167, opened 2016-07-24
Automatic displaying of README.md / README.rst


jelmer:
It would be neat if klaus could automatically display README.md / README.rst files, much like GitHub.

jonashaag:
👍

PR #172, opened 2017-01-26
Add an 'index' page that displays a bit of history and README.

jelmer:
This implements https://github.com/jonashaag/klaus/issues/167.


jonashaag:
So cool! Thanks a lot!

One minor thing though, where would the `/history/` URL be used? I haven't seen it while playing around with klaus; it seems redundant.

jelmer:
(The test failure is python2.6-specific and seems to be in one of the dependencies, unrelated to my change).

The history page is still what's linked in other places - it doesn't display the README doc. The index page also links to the history page (without README) if you click on "Next page".

jonashaag:
Yes but it never links to this specific rule `('history',     '/<repo>/history')`, does it?

jelmer:
Ah, I see what you mean. No, that's true.

jonashaag:
Merged, thanks a lot!

PeterDaveHello:
Is it possible to add an option to decide display it or not? This feature is awesome 👍

jonashaag:
Sure, but why would you want that?

PeterDaveHello:
Just be too familiar with the origin style 😅 thanks

jonashaag:
Haha ok, then I suggest you make a fork of klaus and change the `'index'` route to `'history'` everywhere in the code (should be only a handful places).

PR #173, opened 2017-01-29
Support gitweb .git/cloneurl file and 'gitweb.url' settings.

jelmer:
This addresses #122 when the smart server is disabled.

At the moment this displays the URL in addition to the smart server URL if repo.cloneurl is set.


jonashaag:
Thanks!

PR #179, opened 2017-03-22
Avoid rendering, then parsing diff.

### LINKED ISSUE #78: Get rid of difflib.py
http://stackoverflow.com/questions/774316/python-difflib-highlighting-differences-inline/

If that (plus difflib) is equivalent to the code we're using right now (`difflib.py`) then I'm for using that StackOverflow code.

This is very low priority but maybe someone wants to have a little fun here.

jelmer:
Fixes #78

### DIFF
--- a/klaus/diff.py
+++ b/klaus/diff.py
@@ -8,168 +8,71 @@
     :copyright: 2007 by Armin Ronacher.
     :license: BSD
 """
-import re
-from cgi import escape
+
+from difflib import SequenceMatcher
+
+def highlight_line(old_line, new_line):
+    """Highlight inline changes in both lines."""
+    old_line = old_line.rstrip(b'\n')
+    new_line = new_line.rstrip(b'\n')
+    start = 0
+    limit = min(len(old_line), len(new_line))
+    while start < limit and old_line[start] == new_line[start]:
+        start += 1
+    end = -1
+    limit -= start
+    while -end <= limit and old_line[end] == new_line[end]:
+        end -= 1
+    end += 1
+    if start or end:
+        def do(l, tag):
+            last = end + len(l)
+            return b''.join(
+                [l[:start], b'<', tag, b'>', l[start:last], b'</', tag, b'>',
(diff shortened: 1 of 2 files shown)


jonashaag:
Wow, cool! Thanks a lot!

jelmer:
On Wed, Jun 07, 2017 at 09:25:19AM -0700, Jonas Haag wrote:
> jonashaag commented on this pull request.
> 
> @jelmer we both seem to have forgotten about this, sorry for that. What do you think about the copyright thing?
> 
> > +                'no_newline': (line[-1:] != b'\n')
> +            })
> +            return lines[-1]
> +        chunks.append(lines)
> +        for tag, i1, i2, j1, j2 in group:
> +            if tag == 'equal':
> +                for c, line in enumerate(a[i1:i2]):
> +                   add_line(i1+c, j1+c, 'unmod', line)
> +            elif tag == 'insert':
> +                for c, line in enumerate(b[j1:j2]):
> +                   add_line(None, j1+c, 'add', line)
> +            elif tag == 'delete':
> +                for c, line in enumerate(a[i1:i2]):
> +                   add_line(i1+c, None, 'del', line)
> +            elif tag == 'replace':
> +		# TODO: not sure if this is the best way to deal with replace
> 
> If you think it's better some other way, feel free to change it. Not sure if this was a deliberate decision.
I think the current behaviour from the patch is reasonable (it matches
what currently happens), showing a replacement as delete + add. As
a follow-up, it might be nice to show replacements with a different
color or something like that.

> > @@ -8,168 +8,78 @@
>      :copyright: 2007 by Armin Ronacher.
>      :license: BSD
>  """
> Can we do aside with the copyright notice, or does this still use some of Armin's code?
It's based on it, so it might be prudent to keep the copyright notice
(but IANAL, etc etc).

jelmer:
I've rebased and fixed the last test; this should now be ready.

jonashaag:
Thanks a million!

PR #182, opened 2017-04-28
load line number of the tag in the tagsfile

flares:
As per the documentation of ctags (http://ctags.sourceforge.net/ctags.html), we need to specify ctags fields +n to get the line number of the tags.


jonashaag:
Sounds reasonable. I wonder why it worked for me without the `-n` option?

Issue #36, opened 2012-08-13
Ref names may not contain /


jonashaag:
`feature/foo` doesn't work. Bummer.

jonashaag:
This is ready for implementation, see comments in #103

stpierre:
I'm willing to work on this, but I'm not sure what the final conclusion was. #103 makes two suggestions -- URL encoding, and using : as a separator -- and it's not clear which is preferred. A comment in this thread is pretty disdainful of using : as a separator.

One advantage of using : as a separator is that it will work with Gerrit, which is actually my specific use case. Gerrit can integrate with a git browser, but its integration isn't smart enough to be able to URL encode only certain parts of the URL. (That is, you can't tell Gerrit to URL encode the branch name, but not the file path.) So between that, and the ease of implementing : as a separator, I'd lean towards that.

If I can get clear guidance on what the desired implementation is, and if it will work with Gerrit, I'll make it happen.

jonashaag:
At this point I'd actually prefer the GitHub way, that is, doing manual routing here.

So `foo/bar/spam/eggs` could mean `(foo, /bar/spam/eggs)`, `(foo/bar, /spam/eggs)`, `foo/bar/spam, /eggs)` from klaus' point of view.

I'm not sure how difficult this is to implement though -- IIRC the implementation of @posativ in #103 does exactly this. Maybe it is a good starting point.

If that's problematic to implement, I'd go with the `:` suggestion, that is:

- If there's a `:` in the path/branch string, split at the leftmost `:`.
- Otherwise split at the leftmost `/`
- Generate `:` if and only if necessary

PR #183, opened 2017-06-07
Handle refs that contain a / properly

stpierre:
This fixes two bugs:

* #36 is fixed by forward-porting closed PR #103 
* The drop-down list of branches doesn't work on the index page because query string parameters aren't parsed; this adds parsing so that the rev can be specified in the query string.

### DIFF
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -60,6 +60,7 @@ class Klaus(flask.Flask):
             ('patch',       '/<repo>/commit/<rev>.diff'),
             ('patch',       '/<repo>/commit/<rev>.patch'),
             ('index',       '/<repo>/'),
+            ('index',       '/<repo>/<path:rev>'),
             ('history',     '/<repo>/tree/<rev>/'),
             ('history',     '/<repo>/tree/<rev>/<path:path>'),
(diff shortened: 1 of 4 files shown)


jonashaag:
Thanks a lot!

Why do we even look at the query parameters for the rev? (From a quick look at the source code, I couldn't even find any code that deals with query parameters -- it's been a long time since I worked on klaus... do you know where query parameters are dealt with?)

stpierre:
It's in the branch selector dropdown: https://github.com/jonashaag/klaus/blob/master/klaus/templates/base.html#L34

If the current view is one that doesn't have path elements for rev and path (e.g., the index view), then they're passed as query parameters.

jonashaag:
Ah, I see. Didn't know Flask automatically adds missing parameters to the query string.

This is definitely an oversight -- we should make the `index` URL understand the `rev` component (path does not make sense there)

stpierre:
Okay, I've updated my PR to handle that issue properly. A much easier fix, too!

jonashaag:
Awesome! Thanks a lot!

posativ:
Happy to see my PR finally merged :)

jonashaag:
@posativ Yeah, sorry for the mess.

PR #187, opened 2017-06-21
Fix bug when the latest ref on a repo is a tag

stpierre:
Tags do not have commit times, but they do appear in the list of
refs. This makes klaus use the latest ref that has a commit_time when
finding the last updated time for a repo.

The easiest way to replicate this is with a repo that contains no commits, only a tag.


jonashaag:
Thanks a lot for the patch!

(How is this even useful, only having a tag but no commit?!)

stpierre:
I'm not sure it is useful, but it happened to us and broke the klaus index page. :)

PR #191, opened 2017-06-26
Fix an issue which led to broken link to repo root.

### LINKED ISSUE #190: Broken link to the root of the repo (version 1.2.0)
Hi,

I have an issue with the links to the root folder of my repos. E.g. on the page of the subfolder "multimedia":
http://git.rybalkin.org/kiin-repo/tree/master/multimedia
The link ". ." is
http://git.rybalkin.org/kiin-repo/tree/master/
which when opened tells us:

`Not Found`
`No such commit 'tree/master/'"`

The culprit seems to be the trailing "/" because the link
http://git.rybalkin.org/kiin-repo/tree/master
works as expected.

At the same time ". ." link from a subfolder of a subfolder, e.g. from here
http://git.rybalkin.org/kiin-repo/tree/master/multimedia/alsa-lib
works as expected (and has no trailing "/").

alekseyrybalkin:
Fixes #190. In the case of path == '' url_for helper appends erroneous
slash at the end of the url. Fix this by setting path to None for root
of the repository.


jonashaag:
Hey, thanks for the patch. `parent_directory` does not look like the correct place for a fix, as it is used in multiple contexts.

I guess it has to be fixed somewhere where `url_for` is called instead. Do you know which `url_for` call results in the wrong URL?

alekseyrybalkin:
Yes, it is `url_for` for directories in `tree.inc.html`. I updated the patch to fix this locally in `tree.inc.html`.

jonashaag:
Ah, I see.

This is actually a common issue in the klaus source code:

```
klaus/templates/base.html
34:      <li><a href="{{ url_for(view, repo=repo.name, rev=branch, path=(path or None)) }}">{{ branch }}</a></li>
40:      <li><a href="{{ url_for(view, repo=repo.name, rev=tag, path=(path or None)) }}">{{ tag }}</a></li>
```

See the `path or None`.

Maybe it the better solution would be to allow `/` at the end of the path instead? Then we wouldn't need these hacks.

alekseyrybalkin:
Good idea. I updated the patch accordingly.

jonashaag:
Perfect, thank you!

PR #196, opened 2017-06-29
Drop argparse requirement (included in Python 2.7+)

sbraz:
HI, I guess this isn't needed since https://github.com/jonashaag/klaus/commit/7cc4c13a95ac9ea023e1ab2a383bfb38098bb0e1


jonashaag:
Thanks!

PR #201, opened 2017-07-27
Stop appending "No new line at end of file" warning for each changed …

alekseyrybalkin:
…line of a diff

### DIFF
--- a/klaus/diff.py
+++ b/klaus/diff.py
@@ -15,8 +15,6 @@ from klaus.utils import escape_html as e
 
 def highlight_line(old_line, new_line):
     """Highlight inline changes in both lines."""
-    old_line = old_line.rstrip(b'\n')
-    new_line = new_line.rstrip(b'\n')
     start = 0
     limit = min(len(old_line), len(new_line))
     while start < limit and old_line[start] == new_line[start]:


jonashaag:
This is on purpose, but I'm happy to discuss other visualisations of that information, maybe with a return symbol and a tooltip, see: https://github.com/yiisoft/yii2/pull/122/files

alekseyrybalkin:
The problem is this icon shows for _every_ changed line (old and new), even if the filea ctually _has_ a new line at the end of file. E.g.: http://git.rybalkin.org/kiin-repo/commit/011be0cd1124155650f5b3d335d8f2a8441c61c9/

After this patch, the icon shows only if the end of file is in the diff and if there is actually no new line at the end of file.

jonashaag:
Oh, I see. Thanks!

jonashaag:
Your patch wasn't correct, I fixed it along with another TODO.

alekseyrybalkin:
Great, thanks!

sbraz:
@jonashaag is this supposed to remove the pretty inline highlights? `highlight_line` is not called any more, sounds like a regression to me.

jonashaag:
Whoops, yes.

PR #208, opened 2018-03-01
add an API to disable highlighting by file name

haldean:
This adds an API that a uWSGI script can use to enable/disable highlighting on a per-file basis. I wanted this for my site because I don't like the syntax highlighting that Pygments uses for .txt files and README files, but it seems like it could be a generally-useful API.

I'm not a huge fan of the name of the `highlight_enabled` parameter, but I couldn't think of a better one. Happy to rename it if you can think of one!

### DIFF
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -15,13 +15,15 @@ class Klaus(flask.Flask):
         'undefined': jinja2.StrictUndefined
     }
 
-    def __init__(self, repo_paths, site_name, use_smarthttp, ctags_policy='none'):
+    def __init__(self, repo_paths, site_name, use_smarthttp, ctags_policy='none',
+                 highlight_enabled=lambda f: True):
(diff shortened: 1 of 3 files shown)


jonashaag:
Thanks for the patch!

What do you think about a small redesign:

- a method `Klaus.should_highlight(filename, data) -> bool` that is passed the filename + the contents
- rename `render_markup` to `enable_markup`
- add `enable_highlight : bool` to the `highlight_or_render` function (not as a function but a boolean)

Also, in your current implementation, `highlight_enabled` can also be used to disable markup rendering. If this is intended, then the name should be changed to reflect this behaviour. If not the logic should be changed accordingly.

As a last step we would have to add a new argument to `make_app` that allows you to give your own `Klaus` subclass, i.e. `make_app(app_cls=MyKlaus)`.

PR #210, opened 2018-06-11
add jupyter notebook loader and custom template

lzfxxx:


klaus/markup.py
```diff
@@ -51,5 +51,20 @@ def render_rest(content):
...
+def _load_notebook():
+    try:
+        import nbformat
+        import nbconvert
+    except ImportError:
+        return
+
+    def render_notebook(content):
+        nb = nbformat.reads(content, nbformat.NO_CONVERT)
+        (output, resources) = nbconvert.HTMLExporter(template_file='./klaus/templates/my_full.tpl').from_notebook_node(nb)
```
jonashaag: `my_full`? Confusing name :-D

jonashaag:
Great idea, thanks for the patch!

I'm worried about security though; does this execute any code on server side? Or is this view only?

klaus/markup.py
lzfxxx: It's just a little CSS modification of [nbconvert template 'full'](https://github.com/jupyter/nbconvert/blob/d35c5155a597766b0ff74f930d0f39908f193f86/nbconvert/templates/html/full.tpl), cause the `width` in `@media` setting of that template will make the html overflow the markup `div`. The naming is not very clear, do you have any suggestions?

lzfxxx:
@jonashaag I think it's view only, to render the code and output in the *.ipynb (It's json format inside).

klaus/markup.py
jonashaag: I see! In this case could we simply use the default template and add the CSS fixes to klaus' CSS file? Or if that doesn't work, can we use inheritance to override the CSS instead of copy-pasting the whole file?

klaus/__init__.py
```diff
@@ -17,7 +17,12 @@ class Klaus(flask.Flask):
 
     def __init__(self, repo_paths, site_name, use_smarthttp, ctags_policy='none'):
         """(See `make_app` for parameter descriptions.)"""
-        repo_objs = [FancyRepo(path) for path in repo_paths]
+
+        user_dirs = repo_paths
+        repo_objs = []
```
jonashaag: These changes seem unrelated

jonashaag:
In terms of security, [XSS problems](https://github.com/jupyter/nbconvert/issues/238) are possible at the least. So we have to account for that. Two ideas:

- Apply a very restrictive Content Security Policy for the whole klaus application, and have 2 versions of .ipynb file viewer: a normal CSP-enabled version, and a a CSP-disabled version. We could toggle between them with a simple URL param e.g. `?nocsp`.
- Maybe we can use this [`SanitizeHTML` preprocessor](https://github.com/jupyter/nbconvert/pull/172), haven't looked at it closely. Also we could have two versions of the viewer page, one which strips all `<script>`, `<style>`, `<object>` etc tags, and another one where everything is allowed.

klaus/__init__.py
lzfxxx: Sorry! These changes are for my self usage. Didn't notice these commits will be related to this pull request. I've checkout a new branch and made a new pull request. It was my first time doing pull request, don't know if it's the right way 😃

lzfxxx: I'll move your comments to the new pull request

lzfxxx: #211 new request, I've move some of your comments to there

PR #211, opened 2018-06-12
add notebook loader and custom template

lzfxxx:


lzfxxx:
Related reviews are in origin pull request #210 (which mixed some unrelated changes, so I closed it)

klaus/markup.py
```diff
@@ -51,5 +51,20 @@ def render_rest(content):
...
+def _load_notebook():
+    try:
+        import nbformat
+        import nbconvert
+    except ImportError:
+        return
+
+    def render_notebook(content):
+        nb = nbformat.reads(content, nbformat.NO_CONVERT)
+        (output, resources) = nbconvert.HTMLExporter(template_file='./klaus/templates/my_full.tpl').from_notebook_node(nb)
```
lzfxxx: Moved from #210
jonashaag: my_full? Confusing name :-D

lzfxxx: It's just a little CSS modification of nbconvert template 'full', cause the width in @media setting of that template will make the html overflow the markup div. The naming is not very clear, do you have any suggestions?

jonashaag: I see! In this case could we simply use the default template and add the CSS fixes to klaus' CSS file? Or if that doesn't work, can we use inheritance to override the CSS instead of copy-pasting the whole file?

lzfxxx: fixed

PR #219, opened 2018-08-20
Recognize more Markdown file extensions for READMEs.

cpressey:
May I propose this one-line change?

The Principled Justification: [elsewhere in the source code](https://github.com/jonashaag/klaus/blob/876ba7536a06c9040e0057ba4b29e0d3d4fb51d2/klaus/markup.py#L33), files with these extensions are treated as Markdown, so they should here too, for consistency.

The Pragmatic Argument: GitHub recognizes these files as READMEs written in Markdown, so klaus should too.


jonashaag:
Thanks!

PR #221, opened 2018-09-25
Don't mangle repository name when path contains ".git"

infertux:
Before this commit, the path `/x/klaus.github.io.git` would return `klaushub.io` instead of `klaus.github.io`.

Thank you for building and maintaining this nice Git web interface. I found it through [this page](https://git.wiki.kernel.org/index.php/Interfaces,_frontends,_and_tools#Web_Interfaces) and sadly many of them are unmaintained...

### DIFF
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -1,5 +1,6 @@
 import os
 import io
+import re
 import stat
 import subprocess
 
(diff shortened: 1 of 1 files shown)


jonashaag:
I think you’re using strip incorrectly here: the passed string serves as a set of chars to strip, not as a suffix. Use rsplit(,1)

infertux:
I'm sorry, my Python is pretty rusty. I force-pushed a new commit using `rsplit()` and `re.sub()`.

jonashaag:
Please try this

infertux:
@jonashaag There is a typo in https://github.com/jonashaag/klaus/commit/f0d206cdb3dabe8ec346e6b567c1fa6f980ce470. It should be `path = path[:-4]` instead of `path = path[-4:]`.

jonashaag:
Oops. Fixed.

infertux:
@jonashaag Yup it's working fine now. Thank you.

PR #225, opened 2018-11-05
fix display of code blocks in index view

royneary:
If readme files contain markup, they weren't displayed correctly in the
index view (no padding, no grey box).

The fix contains two changes:
- Use `force_unicode` on the readme's filename
- use the CSS properties for both blobview and non-blobview

### DIFF
--- a/klaus/views.py
+++ b/klaus/views.py
@@ -278,12 +278,11 @@ class IndexView(TreeViewMixin, BaseRepoView):
                 'rendered_code': None,
             })
         else:
+            readme_filename = force_unicode(readme_filename) 
+            readme_data = force_unicode(readme_data)
             self.context.update({
-                'is_markup': markup.can_render(force_unicode(readme_filename)),
-                'rendered_code': highlight_or_render(
-                    force_unicode(readme_data),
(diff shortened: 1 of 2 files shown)


royneary:
- On the left: The klaus.git index page without the fix (code boxes don't have padding, borders and background):
- On the right: The same page with the fix

klaus

klaus/views.py
```diff
@@ -274,16 +274,15 @@ def make_template_context(self, *args):
             (readme_filename, readme_data) = self._get_readme()
         except KeyError:
             self.context.update({
-                'is_markup': None,
+                'is_markup': False,
```
jonashaag: Why this change?

royneary: It was an experiment before I found the fix. So the change is not necessary but I left it there because `is_markup` sounds like a boolean (what would it mean if it was None?). But AFAIK None and False are equivalient in boolean expressions. You decide.

jonashaag:
Thanks!

jonashaag:
IIRC the None value there actually had some semantic (at least readability) meaning, so I've reverted the change.

PR #227, opened 2018-11-07
fix more css for markup in index view

royneary:
Like #225 but for table and img. They should look the same in index view and blobview.


jonashaag:
:heart:

jonashaag:
Btw I'm very open to any other CSS/UI improvements, including a complete rewrite. I'm a terrible designer.

PR #229, opened 2018-11-10
Try a bit harder to come up with sensible names for tarballs and prefixes in tarballs.

jelmer:
Try a bit harder to come up with sensible names for tarballs and prefixes in tarballs.

For revisions without a tag associated, keep the current name, i.e.:

SHA => klaus@SHA

For tags, cope with common syntaxes for tags:

0.1 => klaus-0.1
v0.1 => klaus-0.1
klaus-0.1 => klaus-0.1

### DIFF
--- a/klaus/utils.py
+++ b/klaus/utils.py
@@ -1,4 +1,5 @@
 # encoding: utf-8
+import binascii
 import os
 import re
 import time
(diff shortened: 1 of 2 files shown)


jonashaag:
Great idea. The code base uses „rev“ for what you called „committish“, I guess it makes sense to keep that consistent (either way).

Do you know what heuristic other  tools use? IIRC the initial tarball naming was inspired by gitweb.

jelmer:
I've updated the code to use 'rev' rather than 'committish' instead.

The heuristics are based on the patterns I've observed packaging various kinds of repositories in Debian. 

GitHub strips of the leading "v" if it's present. GitWeb and GitLab appear to just translate slashes into dashes.

jonashaag:
Please review my changes, I did some refactoring and accounted for tag names that start with 'v'.

jonashaag:
Thanks for your work!

PR #239, opened 2019-05-11
When determining last update to a repository, ignore refs that point at missing objects.

jelmer:
When determining last update to a repository, ignore refs that point at missing objects.

This shouldn't normally happen, but when it does, it should not make klaus
crash altogether.

Both C Git and Dulwich will print these tags but print warnings about them.

### DIFF
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -100,7 +100,12 @@ class FancyRepo(dulwich.repo.Repo):
         time they have been committed to last.
         """
         def get_commit_time(refname):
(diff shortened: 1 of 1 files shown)


jonashaag:
❤️ LGTM could you resolve the merge conflicts?

jelmer:
Done.

jonashaag:
Sorry to resurrect this old PR, just wondering: the KeyError we're catching here, shouldn't it be impossible since it's already checked for in `get_refs()` -> `as_dict()`?

Issue #231, opened 2018-12-21
empty directory cause internal error


WestonLu:
I got a lot of git project under the KLAUS_REPOS_ROOT, but when I add an empty directory 
 in KLAUS_REPOS_ROOT by mistake,  it will cause an internal error and the services are all failed. Is there any option to set so i can make klaus just ignore the bad project so other projects can not be affected? Thank you.

jonashaag:
Maybe you can fix the issue yourself? Patch welcome.

xNephe:
Hi,

I just got the same issue, here is the result of some investigation:

When creating the app, all folder in KLAUS_REPOS_ROOT are collected as parameter:
https://github.com/jonashaag/klaus/blob/62d08c3de14e8992b8a6fca4f572a1354fdc2dbf/klaus/contrib/wsgi_autoreloading.py#L40

Klaus tries to create a dulwich repo for every folder in parameter:
https://github.com/jonashaag/klaus/blob/62d08c3de14e8992b8a6fca4f572a1354fdc2dbf/klaus/__init__.py#L20

Dulwich raise an exception when the folder does not correspond to a git repository:
https://github.com/dulwich/dulwich/blob/458bbb4bdefa9d3fcfbb266f74dc7014ce1fd490/dulwich/repo.py#L871

This exception is currently not handled and cause the internal error.

The error could be fix by handling the exception or by validating the path before, it mainly depends on the vision @jonashaag has about klaus: should it crash when it receive an invalid parameter? Should it try to handle it gracefully?

I will try to propose a fix for the issue if I find some time to set up a dev environment for klaus.

jonashaag:
Thanks for investigating!

Actually I think it's fine to simply ignore failing directories (with a warning on the command line) for the autoreloader. But only for the autoreloader, so that if you invoke klaus using the normal command line it crashes for failing directories.

We can either check directories before passing them to `make_app` (by directly invoking Dulwich), which is easy but maybe not the most elegant. (I really don't care about the elegance bit for the autoreloading code.) Or we can have an option to `make_app` to somehow return failing repositories so we can warn on the command line, or something along these lines. That probably requires some minor refactoring though.

xNephe:
Thanks for the answer! I just saw the similar issue #176, and I added a proposition there (ignore hidden folders).
I don't really like calling dullwhich from the autoreloading code. Maybe having an option `fail on invalid repository`?

jonashaag:
[...]

Issue #176, opened 2017-03-12
Ignore invalid repositories (NotGitRepository: No git repository was found at <path>)


NaoPross:
Hello, I was trying to use Klaus under Alpine Linux but the server crashes because there are other files in the `KLAUS_REPOS_ROOT` path, and I couldn't remove them since they are autogenerated by Finder from Mac OS X (the folder is a samba share on a Mac mounted under Linux with cifs).

Right now I've bodged a solution by modifying `klaus/__init__.py` like so to ignore everything but `*.git` directories but maybe it would be better to implement something like a list of ignored files or to catch the exception.
```
@@ -17,7 +17,7 @@
 
     def __init__(self, repo_paths, site_name, use_smarthttp, ctags_policy='none'):
         """(See `make_app` for parameter descriptions.)"""
-        repo_objs = [FancyRepo(path) for path in repo_paths]
+        repo_objs = [FancyRepo(path) for path in repo_paths if path.endswith(".git")]
         self.repos = dict((repo.name, repo) for repo in repo_objs)
         self.site_name = site_name
         self.use_smarthttp = use_smarthttp

```

Traceback:
```
alpinevm:~# /etc/local.d/gitwebserver.start 
+ gunicorn --env KLAUS_SITE_NAME=Git Server --env KLAUS_REPOS_ROOT=/srv/git/repos -b 0.0.0.0:80 klaus.contrib.wsgi_autoreload
[2017-03-12 21:48:41 +0000] [2197] [INFO] Starting gunicorn 19.6.0
[2017-03-12 21:48:41 +0000] [2197] [INFO] Listening at: http://0.0.0.0:80 (2197)
[2017-03-12 21:48:41 +0000] [2197] [INFO] Using worker: sync
[2017-03-12 21:48:41 +0000] [2200] [INFO] Booting worker with pid: 2200
Reloading repository list...
[2017-03-12 21:48:42 +0000] [2200] [ERROR] Error handling request /
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/gunicorn/workers/sync.py", line 135, in handle
    self.handle_request(listener, req, client, addr)
  File "/usr/lib/python2.7/site-packages/gunicorn/workers/sync.py", line 176, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/usr/lib/python2.7/site-packages/klaus/contrib/wsgi_autoreload.py", line 44, in app
    *args, **kwargs
[...]

PR #240, opened 2019-06-12
Handling invalid git repository and hidden folder

xNephe:
This avoid getting an error 500 on an invalid git repository, as discussed in #176 and #231.

- the autoreloading app ignores hidden folders in `KLAUS_REPOS_ROOT`

- invalid repository are displayed in red in the repos list alongside an error message, instead of triggering an error 500.

The index page with invalid repos look like this:
2019-06-11-185356_530x466_scrot

I am terrible at designing things, so advice to improve the display of invalid repositories are welcome!

### DIFF
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -2,6 +2,7 @@ import jinja2
 import flask
 import httpauth
 import dulwich.web
+from dulwich.errors import NotGitRepository
 from klaus import views, utils
 from klaus.repo import FancyRepo
 
@@ -17,8 +18,9 @@ class Klaus(flask.Flask):
 
     def __init__(self, repo_paths, site_name, use_smarthttp, ctags_policy='none'):
         """(See `make_app` for parameter descriptions.)"""
-        repo_objs = [FancyRepo(path) for path in repo_paths]
+        repo_objs, invalid_repos = self.load_repos(repo_paths)
         self.repos = dict((repo.name, repo) for repo in repo_objs)
+        self.invalid_repos = invalid_repos
         self.site_name = site_name
         self.use_smarthttp = use_smarthttp
         self.ctags_policy = ctags_policy
@@ -79,6 +81,15 @@ class Klaus(flask.Flask):
         else:
             raise ValueError("Unknown ctags policy %r" % self.ctags_policy)
 
+    def load_repos(self, repo_paths):
+        repo_objs = []
+        invalid_repos = []
+        for path in repo_paths:
+            try:
+                repo_objs.append(FancyRepo(path))
+            except NotGitRepository:
+                invalid_repos.append(path)
+        return repo_objs, invalid_repos
 
 
 def make_app(repo_paths, site_name, use_smarthttp=False, htdigest_file=None,
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -4,7 +4,7 @@ import httpauth
 import dulwich.web
 from dulwich.errors import NotGitRepository
 from klaus import views, utils
-from klaus.repo import FancyRepo
+from klaus.repo import FancyRepo, InvalidRepo
 
 
 KLAUS_VERSION = utils.guess_git_revision() or '1.4.0'
@@ -88,7 +88,7 @@ class Klaus(flask.Flask):
             try:
                 repo_objs.append(FancyRepo(path))
             except NotGitRepository:
-                invalid_repos.append(path)
+                invalid_repos.append(InvalidRepo(path))
         return repo_objs, invalid_repos
 
 
--- a/klaus/templates/repo_list.html
+++ b/klaus/templates/repo_list.html
@@ -36,7 +36,7 @@
   {% endfor %}
 </ul>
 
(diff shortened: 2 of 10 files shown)


jelmer:
FWIW the code looks reasonable to me; the tests all pass except for python 3.7 (fixed in master).

klaus/templates/repo_list.html
```diff
@@ -36,4 +36,17 @@ <h2>
   {% endfor %}
 </ul>
 
+{% if invalid_repos is defined and invalid_repos|length>0 %}
```
jonashaag: Is it ever undefined?

klaus/__init__.py
```diff
@@ -17,8 +18,9 @@ class Klaus(flask.Flask):
 
     def __init__(self, repo_paths, site_name, use_smarthttp, ctags_policy='none'):
         """(See `make_app` for parameter descriptions.)"""
-        repo_objs = [FancyRepo(path) for path in repo_paths]
+        repo_objs, invalid_repos = self.load_repos(repo_paths)
         self.repos = dict((repo.name, repo) for repo in repo_objs)
+        self.invalid_repos = invalid_repos
```
jonashaag: Shouldn't we also use `.name` for invalid repos? It's based on the path only, so you could factor it out from `Repo.name` to e.g. `utils.something` and use it here as well.

klaus/views.py
```diff
@@ -37,7 +37,8 @@ def repo_list():
     else:
         sort_key = lambda repo: (-(repo.get_last_updated_at() or -1), repo.name)
     repos = sorted(current_app.repos.values(), key=sort_key)
-    return render_template('repo_list.html', repos=repos, base_href=None)
+    invalid_repos = sorted([os.path.basename(path) for path in current_app.invalid_repos])
```
jonashaag: See comment about `.name` above.

Maybe it's handy to use a `InvalidRepo` class that has a `.name` attribute. So `FancyRepo` for good repos, `InvalidRepo` for bad repos. Not sure.

tests/repos/scripts/hidden_folder
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+# Hide the directory
+cd ..
+mv hidden_folder .hidden_folder
```
jonashaag: Looks like some changes to the actual tests are missing from the commit?

xNephe: No, this file add a hidden folder in `KLAUS_REPOS_ROOT`. The already existing test `test_wsgi_autoreload` will only pass if it correctly ignores hidden folders, but no new test case is needed for this feature.

klaus/templates/repo_list.html
xNephe: Not in the current code. I will remove the check.

klaus/views.py
xNephe: An `InvalidRepo` class seems coherent to me, especially if more information about invalid repos is stored in the future (the reason why it is invalid for instance). I will add a new class for them.

klaus/__init__.py
xNephe: I am not sure what you mean here. Are you suggesting something like `self.invalid_repos = dict((repo.name, repo) for repo in invalid_repos)`? I don't see the use of a dict here...

tests/repos/scripts/hidden_folder
xNephe: However, I did not add a test for the invalid repo display part, I should create one.

klaus/__init__.py
jonashaag: True, at the moment we don’t need it since we’re not mapping invalid repos to urls. If we do that at some point we can change it to a dict. 

OTOH we could even stuff both types of repo objects into the same dict and have a check whenever we list or show repos.

xNephe: I would rather keep two separated collections. I can't think of an operation we would want to apply to both valid and invalid repositories without distinction (even for display, we always want to separate valid and invalid repositories...). Having them in the same collection would requires to split the collection whenever we iterate through it, which is error prone.

jonashaag:
@xNephe you still working on this?

xNephe:
I didn't find time to work on it in the last weeks, I have been pretty busy, but I haven't forgot it.
I hope to find some time by the end of the week.

xNephe:
I eventually found some time to fix the PR:
- I added an InvalidRepo object to abstract invalid repos
- I added a test to ensure invalid repos are display in the repo list view

jonashaag:
Thanks a lot! Merged with a few minor changes in ee9d78044040b8bd9090b774838a9f6684941f2d and 7f3f5b49990076c0c0e5338a31a523a70c92b750

xNephe:
Thanks! However, I noticed you added a function `is_hidden_folder` that end up not being used, because of the switch from `listdir` to `glob.glob`. We may want to delete it?

jonashaag:
Oops thanks!

rotsix:
Hey, is there a way to hide those invalid repositories?

PR #247, opened 2020-02-27
Fix import of ProxyFix for Werkzeug 1.0.0

embeddedc:
Fix import of ProxyFix for werkzeug 1.0.x

### DIFF
--- a/klaus/utils.py
+++ b/klaus/utils.py
@@ -14,7 +14,7 @@ try:
 except ImportError:
     chardet = None
 
-from werkzeug.contrib.fixers import ProxyFix as WerkzeugProxyFix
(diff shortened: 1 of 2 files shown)


jonashaag:
Is this also compatible with older versions of Werkzeug? Otherwise we’d have to specify minimum version in requirements.

embeddedc:
`werkzeug.middleware.proxy_fix` has been introduced with `werkzeug 0.15.0`, so that would be the minimum version.

`pip install klaus `will currently import `werkzeug 1.0.0`, so everybody will get a `No module named 'werkzeug.contrib'` error

jonashaag:
<3

embeddedc:
Danke nach 
😊

PR #254, opened 2020-07-03
Have ?raw=true in querystring on blob view redirect to raw view.

chaitanyarahalkar:
This will allow images to be rendered in Markdown.

### DIFF
--- a/klaus/views.py
+++ b/klaus/views.py
@@ -2,7 +2,7 @@ from io import BytesIO
 import os
 import sys
 
-from flask import request, render_template, current_app, url_for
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks for the contribution!

Could you please explain in more detail what you’re trying to solve?

Why a redirect? Why not link to the raw file directly?

Also, the change will need to have a unit test.

stappersg:
Title of this _merge request_ inspired me to file [a wish](https://github.com/w4/rgit/issues/25) for another _gitweb server_.

PR #259, opened 2020-10-08
[WIP] PoC for a Node.JS backend using nodegit

julien-c:


node/app/Context.ts
```diff
@@ -0,0 +1,147 @@
...
+		
+		if (this.path === undefined) {
+			throw new NotFoundError(`Invalid blob, path is undefined`);
+		} else {
+			const commitTree = await this.commit.getTree();
+			const treeEntry = await commitTree.getEntry(this.path);
+			if (! treeEntry.isBlob()) {
+				throw new NotFoundError(`No such blob ${this.path} in repository ${this.repoName}/${this.rev}`);
+			}
+			this.blob = await treeEntry.getBlob();
```
Pierrci, line 133: Did you check that this work? Every time I tried `entry.getBlob` in my code I get a `Cannot read property 'repo' of undefined` error, but doing `repo.getBlob(entry.oid())` works 🧐

node/app/Context.ts
```diff
@@ -0,0 +1,147 @@
+import * as express from 'express';
+import * as Git from 'nodegit';
+import { c } from '../lib/Log';
+import { Repo } from './Repo';
+
+
+const DEFAULT_BRANCH = `master`;
```
Pierrci, line 7: `main` or `master`?

julien-c: master here, main at our place I guess!

node/app/Context.ts
julien-c: Yes it works here (AFAIK), though I've seen the error you reference before.

Pierrci: Ok (same for `entry.getTree` for me)

node/app/Repo.ts
```diff
@@ -33,11 +40,10 @@ export namespace Repo {
 	): Promise<number> {
 		const revWalk = repo.createRevWalk();
 		revWalk.push(before.id());
-		let i = 1;
+		let i = 0;
 		/// ^^ Include `before` in the count.
 		while (true) {
 			try {
-				await revWalk.next();
```
Pierrci: Huuuuuuh, sure? 😬

Pierrci: (I think there is now an infinite `while` no?)

julien-c: Weird! Not sure what happened here.

node/README.md
```diff
@@ -0,0 +1,41 @@
...
+Non-bare (regular repos with a working copy) are convenient for testing as you can create a dummy commit history or tree structure to check that everything works fine.
+
+### Step 2:
+
+Run the node app
+
+```
+cd node
+npm i
+tsc
```
jonashaag, line 18: I think this should be `tsc -w`

node/app/routes.ts
```diff
@@ -0,0 +1,276 @@
...
+		}
+	}
+	const allCommits = await Promise.all(oids.map(x => context.repo.getCommit(x)));
+	/// ^^ this is very fast.
+	/// no pagination (we haven't had the use for now.)
+	let commits: Git.Commit[] = [];
+	if (context.path === undefined) {
+		commits = allCommits;
+	} else {
+		/// CAUTION(this can be pretty long)
```
jonashaag, line 241: A simple way to speed this up is by not fetching the entire diff, but comparing the parents' trees to the commits' trees (without actually looking at the file contents). Looking at `.coveragerc` history takes around 30 s for me.

node/server.ts
```diff
@@ -0,0 +1,155 @@
...
+app.get('/:namespace/:repo/blame/:rev/*',   blameBlob);
+
+
+app.get(           '/:repo/commit/:rev',    viewCommit);
+app.get('/:namespace/:repo/commit/:rev',    viewCommit);
+
+
+
+app.get(           '/:repo/commits',        historyCommits);
+app.get('/:namespace/:repo/commits',        historyCommits);
```
jonashaag, line 122: This URL scheme doesn't allow for a repository called "commits" in a namespace, ie. `somenamespace/commits`. That's why I decided to use the `~` prefix in Klaus. Maybe there are other ambiguous names, I haven't checked thoroughly.

node/server.ts
```diff
@@ -0,0 +1,155 @@
...
+	express.static(`${__klausDir}/static`)
+);
+app.get(
+	'/favicon.ico',
+	(req, res) => res.sendFile(`${__klausDir}/static/favicon.png`)
+);
+
+/**
+ * Note: we only support branch names and tag names
+ * not containing a `/`.
```
jonashaag, line 40: I see where you're coming from but at least in the software engineering world feature branches (`feature/xyz`) are very common.

jonashaag:
Reviewed most of the code and also did some performance tests on my machine. Implementation and performance look very good. libgit2 performance is 1–2 order of magnitudes better than Dulwich performance, apparently.

I added a few minor review comments. I don't think there are any obviously bad implementation choices.

I suggest that you performance monitor as much as possible in production: Some of the history-based views are unbounded (eg. counting number of commits) and may turn out to be too slow. It probably also makes sense to add caching here and there if you have an easy way to do cache invalidation (with the current refresh mechanism the page is essentially static so you could just cache everything).

node/app/routes.ts
jonashaag: See https://github.com/jonashaag/klaus/blob/purepy-hist/klaus/repo.py#L199 for an implementation

node/server.ts
julien-c: Yes – for a v1 for the sake of simplicity I think this will work but we can always add support later

node/server.ts
julien-c: Yes, we are kind of constrained in the URLs we have to support (as they already exist) so I'll just hardcode a blacklist of forbidden repo names

node/app/routes.ts
julien-c: `.coveragerc` in which repo?

julien-c: (and 30s vs. how long with this current implem?)

Yes I think you make a good point. Will have to experiment and re-implement.

jonashaag: Sorry, should've been more specific. I'm talking about `huggingface/transformers/commits/master/.coveragerc`. It takes around 30 s, vs `time git log .coveragerc` => 145 ms

jonashaag: In Klaus I implemented it simply by calling `git log` using subprocess. I used the pure Python implementation linked above but it was very slow.

julien-c: Ok, will take a stab at this now (unless you want to do it? :) )

jonashaag: No, go for it... I invested 20 min into coming up with a draft but realised it's going to take me much longer than someone proficient in TypeScript

jonashaag: I also don't know if what I came up with in Klaus is the best way to do it (but it works)

julien-c: Interesting SO question here: https://stackoverflow.com/questions/21178383/how-is-gits-log-filename-implemented-internally

PR #262, opened 2020-10-29
added power support arch on yml file.

srinivas32:
Added power support for the travis.yml file with ppc64le. This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,3 +40,12 @@ jobs:
           tags: true
         provider: script
         script: ./build-and-push-docker.sh
+        # added power support arch.
+    - stage: deploy
+      script: skip
+      arch: ppc64le
+      deploy:
+        on:
+          tags: true
+        provider: script
+        script: ./build-and-push-docker.sh


jonashaag:
Are you a bot?

gerrith3:
@jonashaag no, Srinivas (@srinivas32 ) isn't a bot - he's a contractor working for the IBM Power Brand as we work to consolidate the widely distributed testing by the distros into a more top of tree approach similar to what has been done with Travis CI on Intel.  The goal is to support any projects currently shipped as part of major distributions (e.g. debian/ubuntu, or fedora/RHEL/CentOS) that is using Travis CI for Intel today and extending that yaml file to cover ppc64le testing, which we have recently made available with Travis CI, including free hardware for open source projects.  What makes you question whether he is a bot?

jonashaag:
Ok, thanks! I wondered because he sends the same message and the same changes to a lot of repos.

Question for understanding the PR: What's the point of building the Docker image twice, and also push it twice to Docker Hub? Aren't the images going to be exactly identical no matter the Docker host?

Also, honestly, before today I have never even heard of ppc64le architecture, let alone has support for ppc64le ever been requested by anyone, so I don't know how to feel about supporting such exotic architecture. Are you going to keep Klaus ppc64le support up to date, i.e. fix regressions in the future? Otherwise I don't see how this can work... On the other hand, why test specific CPU architectures for a Python-only project anyways? Klaus does not include any architecture specific code.

gerrith3:
@jonashaag yeah, apologies for the perception - he is just very active/very productive.  A shame that makes him look like a bot.

So, Docker supports something called "multi-arch" or "fat manifests".  Dumbed down version, there is a container for each architecture type (the actual machine code instructions for each architecture are different) and then provides a file which is essentially an intelligent symbolic link that pulls up the right container based on the current architecture the code is running on.  So, containers abstract all of the components you might need on a given architecture, but they don't replace the need for architecture dependent containers.

As far as maintenance, you may have in part answered your own question.  ;)  If indeed it is 100% architecture independent then, no harm done, we have no maintenance, you have no maintenance.  If it is not, then yes, we can help, although with checks like Travis it should make the change immediately obvious even before the submitter gets to  you, meaning that they will see and probably have an idea then that they are making the code architecture dependent and you can decide what you think about that.  BTW, ppc64le is in comparison more similar to Intel than most other architectures (s390x, arm, x86, etc.) so  most things will "just work"  And, before submiting the patch, @srinivas32 did local builds and testing, tested with his own travis account, and then prepared the patch for you.  So, you can be pretty confident that you are indeed, at this moment, architecture independent.  As far as hearing about the architecture, you can probably find it at almost every Global 1000 company - or look to desktop offerings from Raptor.  The majority of all Oracle installations (at least at size) have been on the Power architecture for years.  Check out the top 100 supercomputers, you'll see Power there.  Or SAP is another.  As a niche, we are a multi-billion dollar little niche, I gues.  ;)

I hope that helps a bit?

jonashaag:
Thanks, that was helpful. 

If anyone reads this who is using Klaus on a supercomputer, please let me know! 😁

Will merge this tentatively: if it causes problems that neither of us wants to fix, I’ll remove it.

jelmer:
@gerrith3  Out of curiosity (and as the Debian/Ubuntu maintainer) how does this help with the Ubuntu packaging?

gerrith3:
@jelmer This is a little complicated and some isn't public, but we are swizzling how we make hardware available to everyone, including debian farms, Ubuntu farms and others.  There is a big push to move requests to the cloud over time and reduce the hardware investment in all of the downstream communities which are highly replicated and instead do more validation upstream.  The goal is to reduce duplication and use more of the model with Intel and using CI's to automate testing whenever possible. This is also where arm is kind of going, and s390x is doing the same, albeit a little less broadly.

PR #266, opened 2021-04-22
add 'hide-invalid-repositories' option

rotsix:
Hey.
As I don't mind if Klaus find some invalid repositories, I added an option to hide those.

During the app building, when it formats the {in,}valid repositories, it checks for this newly created option and dismiss the invalid ones.

I hope that's enough for you!
Thanks for this super software :)

### DIFF
--- a/bin/klaus
+++ b/bin/klaus
@@ -63,6 +63,13 @@ def make_parser():
         type=git_repository,
     )
 
+    parser.add_argument(
+        "--hide-invalid-repos",
+        help="hide invalid repositories",
+        default=False,
+        action="store_true",
+    )
+
     grp = parser.add_argument_group("Git Smart HTTP")
     grp.add_argument(
         "--smarthttp", help="enable Git Smart HTTP serving", action="store_true"
@@ -121,6 +128,7 @@ def main():
 
     app = make_app(
         args.repos,
+        args.hide_invalid_repos,
         force_unicode(args.site_name or args.host),
         args.smarthttp,
         args.htdigest,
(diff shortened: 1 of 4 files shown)


jonashaag:
I’m sorry, I think I didn’t respond here because the tests fail and I wanted to wait for you to fix them before reviewing the PR

rotsix:
Hey again, thank you for your time. I believe all is working this time!

jonashaag:
Thanks a lot, please move the new setting to very end so as to make the changes backwards compatible.

Also what do you think about adding a "N invalid repositories hidden" to the repo list page? Do you think that could be helpful? Or maybe we can just log a warning to the terminal. Just a way for people to discover that some stuff has been hidden. Not sure what's the best option here.

jonashaag:
I just tried this, but when I start `klaus --hide-invalid-repos some_invalid_repo`, I still get `klaus: error: argument DIR: 'some_invalid_rep': Not a Git repository`. I was expecting the option to ignore that repo, or did I misunderstand what the option is supposed to do?

rotsix:
Using the standalone application (`python bin/klaus`) won't load an invalid repository and end with an error (`klaus: error: argument DIR: <path>: Not a Git repository`).

By using wrapping Klaus within `wsgi`, I believe you can make it works with some invalid repositories. That's how I use it daily with the Docker image.

jonashaag:
Yes, let's do the check later, for example in `main` where we do the other checks.

rotsix:
Hey @jonashaag, have you had some time to think about those type checks? :)

PR #267, opened 2021-05-02
Add auto_klaus.py to autodetect repositories with low complexity

wsldankers:
Alternative take on the "automatically discovered repositories" concept that requires no threads, polling or inotify. Instead the filesystem is consulted whenever a repository name is looked up.

Since `os.path.exists()` and `os.listdir()` are fairly quick filesystem operations, performance should be good for small to medium sites. `FancyRepo()` objects are cached.

Repositories are identified by the existence of a `<reponame>/git-daemon-export-ok` file (for compatibility with gitweb).

Example usage:

```python
from klaus.contrib.auto_klaus import Klaus

application = Klaus('/srv/git', "My git repositories", False)

application.wsgi_app = httpauth.AlwaysFailingAuthMiddleware(
    wsgi_app=dulwich.web.make_wsgi_chain(
        backend=dulwich.server.DictBackend(SlashDynamicRepos(application.valid_repos)),
        fallback_app=application.wsgi_app,
    ),
)
```


jonashaag:
I like the idea, adding a few code comments.

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
+"""
+Alternative take on the "automatically discovered repositories" concept
+that requires no threads, polling or inotify. Instead the filesystem is
+consulted whenever a repository name is looked up.
```
jonashaag, line 4: I feel like it's important to mention here that each time the repo list is viewed by someone, the following code is run:

- 1x `listdir()` in the root
- Nx `os.path.exists` in each of the folders

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+
+Repositories are identified by the existence of a
+
+    <reponame>/git-daemon-export-ok
+
+file (for compatibility with gitweb).
+
+Example usage:
+
+    from klaus.contrib.auto_klaus import Klaus, SlashDynamicRepos
```
jonashaag, line 18: Maybe we can find a more descriptive name than `auto_klaus`

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+
+    application = Klaus('/srv/git', "My git repositories", False)
+
+    application.wsgi_app = httpauth.AlwaysFailingAuthMiddleware(
```
jonashaag, line 22: Most people using klaus won't be able to understand how to actually run a klaus server with this code. We should have something as simple as this: https://github.com/jonashaag/klaus/wiki/Autoreloader

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+    Maintain a virtual read-only dictionary whose contents represent
+    the presence of git repositories in the given root directory.
+    """
+
+    def __init__(self, root):
+        self._root = pathlib.Path(root)
+        self._repos = {}
+
+    def __getitem__(self, name):
+        if not name or name[0] == '.' or '/' in name or '\0' in name:
```
jonashaag, line 49: I think we should also check for `os.pardir` here?

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+            raise KeyError(name)
+
+        repos = self._repos
```
jonashaag, line 52: That alias is confusing, please don't do it.

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+
+        repo = klaus.repo.FancyRepo(str(path))
+        repos[name] = repo
+        return repo
+
+    def __iter__(self):
+        root = self._root
+        return (
+            repo for repo in os.listdir(root)
+                if os.path.exists(root / repo / 'git-daemon-export-ok')
```
jonashaag, line 71: Could we save a few `exists()` checks here by checking `self._repos` for a positive match first?

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+            raise KeyError(path)
+        return self._base[path[1:]]
+
+    def __iter__(self):
+        return ('/' + name for name in self._base)
+
+    def __len__(self):
+        return len(self._base)
+
+class Klaus(klaus.Klaus):
```
jonashaag, line 97: I don't think it's wise to create a subclass of `klaus.Klaus` because it will potentially break anytime we make changes to `klaus.Klaus`. Can you think of another way of solving this that doesn't require using a "private" API? (I'm happy with doing changes to `klaus.Klaus` if it's necessary.)

klaus/contrib/auto_klaus.py
```diff
@@ -0,0 +1,103 @@
...
+        self._root = pathlib.Path(root)
+        self._repos = {}
+
+    def __getitem__(self, name):
+        if not name or name[0] == '.' or '/' in name or '\0' in name:
+            raise KeyError(name)
+
+        repos = self._repos
+        path = self._root / name
+        if not os.path.exists(path / 'git-daemon-export-ok'):
```
jelmer, line 54: what about non-bare repositories?

klaus/contrib/auto_klaus.py
wsldankers: It's included in `name[0] == '.'` for most common operating systems but we could include `os.pardir` and `os.curdir` explicitly, yes.

klaus/contrib/auto_klaus.py
wsldankers: That's what I had at first, in fact. But then the admin cannot retract access from a repository. I suppose we could make that optional.

klaus/contrib/auto_klaus.py
wsldankers: Well we could change these lines in `klaus/__init__.py` to always use the `SlashDynamicRepos` proxy (perhaps under a different name):
```python3
        dulwich_backend = dulwich.server.DictBackend(
            {
                "/" + namespaced_name: repo
                for namespaced_name, repo in app.valid_repos.items()
            }
        )
```
Then add an optional named parameter so we can pass a constructor that determines how to handle the first argument to `Klaus()`, dependency-injection style. `load_repos()` would be moved to a default implementation for that.

One thing I didn't get though, is why `invalid_repos` exists, it seems like it could be an information leak.

klaus/contrib/auto_klaus.py
wsldankers: I suppose we could make the path to check configurable, so you can configure it to be `.git/git-daemon-export-ok` or even something completely different if you don't care about gitweb conventions.

klaus/contrib/auto_klaus.py
wsldankers: Yeah this was just the first name I came up with as I was throwing this together for my own server, heh. I'm open to suggestions. The two hardest problems in computer science: naming things, cache invalidation and off-by-one errors. 😉

klaus/contrib/auto_klaus.py
wsldankers: Sure, I'll look into that.

klaus/contrib/auto_klaus.py
jelmer: Checking for git-daemon-export-ok is quite specific to being compatible with gitweb, and a change from the behaviour that currently exists in klaus. For example, most of my repositories that are visible in klaus don't have a git-daemon-export-ok file and wouldn't work with this script. Perhaps it would make sense to call that out more explicitly?

klaus/contrib/auto_klaus.py
jonashaag: Right but if we keep the safety check + `.pop()` in `__getitem__` that's an acceptable compromise IMO.

klaus/contrib/auto_klaus.py
jonashaag: I agree that it's kind of annoying that we have multiple ways to deal with autodiscovery/autoreloading. It long overdue someone contributes a well thought out solution into Klaus proper.

I wonder if we should make this "flag file" approach part of the existing autoreloading code? To me it seems the only user-visible difference between this implementation and the existing one the `git-daemon-export-ok` file.

klaus/contrib/auto_klaus.py
wsldankers: I don't follow. The “safety check + `.pop()`” is what we have now. The alternative is to remove the safety check and just rely on the presence of `self._repos[repo]`. I don't see the compromise you're trying to get at.

klaus/contrib/auto_klaus.py
wsldankers: Added a blurb to that effect to the `wsgi_autodetecting.py` docstring.

klaus/contrib/auto_klaus.py
wsldankers: Renamed it to `wsgi_autodetecting.py`. Still a bit generic but hopefully more descriptive.

klaus/contrib/auto_klaus.py
wsldankers: Added a `wsgi_autodetect.py` in the style of `wsgi_autoreload.py`.

klaus/contrib/auto_klaus.py
wsldankers: Added all of these to the validation expression.

klaus/contrib/auto_klaus.py
wsldankers: Removed (and one other instance of the same pattern as well).

Not a fan of the repeated subexpressions it causes, though.

klaus/contrib/auto_klaus.py
wsldankers: Added the `detect_removal` option as a compromise between performance and correctness.

klaus/contrib/auto_klaus.py
wsldankers: Implemented: `Klaus` now has a `repo_container_factory` (apologies for the Java-esque naming) that receives the `repo_paths` parameter as its argument and returns a repository container. Such a container implements the python `dict` interface and additionally has an `invalid` property to store any invalid repositories (exposed as `invalid_repos` on `Klaus`).

ldrumm:
This is a great idea. I'm going to close my 5 year old inotify PR that I was one day going to get round to in favour of this one.

klaus/contrib/auto_klaus.py
ldrumm: I think there's a bug here WRT `.git` suffix handling

If I have a repo named `repo.git` with `repo.git/git-daemon-export-ok`, then it's properly listed on the index page, but the link generated resolves to `/repo` which then 404s because `./repo/git-daemon-export-ok` doesn't exist. `klaus.contrib.autoreload` exhibits the same suffix removal, but correctly resolves the shortened name.

I'm not sure what the best approach to resolve this is, but I think this mapping might need a minor rethingk

ldrumm (commented):
I've tested this on my server and it's performing well modulo the one issue described above. I like the `git-daemon-export-ok` idea too as I'd otherwise create a `public` directory and a bunch of symlinks which I've never entirely conveniced by.

klaus/contrib/auto_klaus.py
wsldankers: The path is now configurable and there's also a way to tell it to simply consider all directories as valid.

wsldankers: I've tried to add functionality to deal with this situation. The `KLAUS_DIRECTORY_SUFFIXES` environment variable can now contain any number of potential directory suffixes (no suffix and `.git` by default).

PR #269, opened 2021-05-19
Check that README is actually a blob, not a tree

jelmer:
Check that README is actually a blob, not a tree.

I did find one upstream package that has a directory called README :)


jonashaag:
Software engineering is all about wrong assumptions 😄

PR #270, opened 2021-06-02
Install python3 plugin for uwsgi

jelmer:
Install python3 plugin for uwsgi.


jonashaag:
Looks good but I wonder why it worked before?

jelmer:
On Wed, Jun 02, 2021 at 12:42:52PM -0700, Jonas Haag wrote:
> Looks good but I wonder why it worked before?
I was specifying --plugin python3 since -w didn't work for me. It does
seem to work fine in debug mode with the current docker image.

-- 
Jelmer Vernooĳ ***@***.***>
PGP Key: https://www.jelmer.uk/D729A457.asc

PR #271, opened 2021-07-14
Added support for serving all repositories in a path.

353fc443:
Previously `klaus` was not able to serve repos in a path 
``` sh 
klaus .
usage: klaus [-h] [--host HOST] [--port PORT] [--site-name SITE_NAME] [--version] [-b] [-B BROWSER]
             [--ctags {none,tags-and-branches,ALL}] [--smarthttp] [--htdigest FILE] [--debug]
             [DIR ...]
klaus: error: argument DIR: '/Users/user': Not a Git repository
```
Now 
```sh
klaus --path .
```
Can serve all git repos present in PWD/The given path

Thank you :)

### DIFF
--- a/bin/klaus
+++ b/bin/klaus
@@ -63,6 +63,12 @@ def make_parser():
         type=git_repository,
     )
 
+    parser.add_argument(
+        "--path",
+        help="serve all repositories in the path",
+        default=None,
+    )
+
     grp = parser.add_argument_group("Git Smart HTTP")
     grp.add_argument(
         "--smarthttp", help="enable Git Smart HTTP serving", action="store_true"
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks for the patch! Did you try https://github.com/jonashaag/klaus/wiki/Autoreloader?

353fc443:
Oh this seems to work with `gunicorn` but for my use case I can't use `gunicorn`.

If you are not happy with this PR then you may close it :)

jonashaag:
You can also make it work with any other WSGI server. What's your setup?

353fc443:
I dont use WSGI as I'm not using for production

jonashaag:
It actually starts a simple WSGI server. It's true that the script doesn't support the autoreloader. I'll be closing this as to not have *another* competing implementation of serving multiple repositories.

PR #274, opened 2022-01-08
Default to running Klaus in Python3

jelmer:
Default to running Klaus in Python3.

### DIFF
--- a/bin/klaus
+++ b/bin/klaus
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # coding: utf-8
 from __future__ import print_function
 


jonashaag:
Hmmm shouldn't `python` be an alias to `python3` on modern systems?

Can we be sure that `python3` exists?

jelmer:
All python 3 installations provide a python3 binary - but on e.g. the last release of Debian python often still points at python2.

This PR is motivated by the fact that I keep invoking bin/klaus on my Debian Sid machine and having it fail because newer versions of dulwich are python3 only.

jelmer:
See also https://www.python.org/dev/peps/pep-0394/

jonashaag:
Thanks!

PR #275, opened 2022-01-08
Really basic implementation of a metrics target for use by prometheus

jelmer:
Really basic implementation of a metrics target for use by prometheus.

Some caveats:

 * this exposes /metrics on the same port as the rest of klaus;
   one might not want to expose internal metrics there
 * this doesn't yet add any really useful metrics
   (number of requests, failures, exceptions, etc). At the moment
   it just exports a bunch of python statstics.

### DIFF
--- a/klaus/__init__.py
+++ b/klaus/__init__.py
@@ -60,7 +60,6 @@ class Klaus(flask.Flask):
         # fmt: off
         for endpoint, rule in [
             ('repo_list',   '/'),
-            ('robots_txt',  '/robots.txt/'),
             ('blob',        '/<repo>/blob/'),
             ('blob',        '/<repo>/blob/<rev>/<path:path>'),
             ('blame',       '/<repo>/blame/'),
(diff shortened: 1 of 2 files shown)


jonashaag:
I wonder if this could be implemented as a WSGI middleware? (Maybe there already exists some pluggable WSGI middleware that we can just use?)

jonashaag:
Maybe one of those, didn't look into any closely though

https://github.com/benley/python-wsgi-prometheus
https://github.com/prometheus/client_python#wsgi
https://github.com/rycus86/prometheus_flask_exporter

jelmer:
It can't be just middleware if you want to add klaus-specific metrics, and those are the most useful.

jonashaag:
Sure, at least we will have to provide some hooks in the Klaus codebase or we can push metrics from Klaus (eg. increase some counters from Klaus views).

I was wondering more about the metrics endpoint. I wonder if it makes sense to use Prometheus-Flask-Exporter or something like this: https://github.com/prometheus/client_python#flask

But I am fine with a stand-alone implementation if you think that’s easier. I wonder if we should make the feature optional even if prometheus_client is installed.

jelmer:
Maybe the best approach here is to have a separate port for the metrics. That's consistent with what other web apps do - so you don't have your metrics exposed to the internet. The other benefit is that we can then only import the metrics library when --metrics-port is specified.

PR #280, opened 2022-06-23
Show READMEs in subdirectories

### LINKED ISSUE #250: Show README files of subfolders
Some repos have README files in each subfolder. Change the commit history view so that it always shows README contents (not only for the / folder)

s-ol:
Close #250

### DIFF
--- a/klaus/views.py
+++ b/klaus/views.py
@@ -224,9 +224,44 @@ class TreeViewMixin(object):
             root_directory = parent_directory(root_directory)
         return root_directory
 
+class ReadmeMixin(object):
+    """The logic required for finding and displaying README files."""
 
-class HistoryView(TreeViewMixin, BaseRepoView):
-    """Show commits of a branch + path, just like `git log`. With pagination."""
+    def _get_readme(self):
+        commit, path = self.context["commit"], self.context["path"]
+        tree = self.context["repo"].get_blob_or_tree(commit, path)
+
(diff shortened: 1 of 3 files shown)


klaus/views.py
```diff
@@ -224,9 +224,44 @@ def get_root_directory(self):
...
+        else:
+            raise KeyError
+
+    def get_readme_context(self):
+        try:
+            (readme_filename, readme_data) = self._get_readme()
+        except KeyError:
+            return {
+                "is_markup": None,
+                "rendered_code": None,
```
jonashaag, line 250: Does it make sense to rename those to something `readme_is_markup` and `readme_is_rendered_code`?

jonashaag (commented):
Great work, thanks a bunch!

klaus/views.py
s-ol: Sure, I found these confusing but left them as is. The second isn't a flag but the rendered result, so I'd go for `readme_content` and `readme_is_markup`, wdyt?

s-ol: Ah, there is no `readme` prefix because the template is actually shared with the [`FileView`](https://github.com/jonashaag/klaus/pull/280/files/678b40231a7b2b24c1d307670a36f7a8b39947ea#diff-6cca1e1a85b1376046f3e0933db6e90785e0ad20a834e8d5536ae295347bd014L470-L476). Merge as-is?

PR #281, opened 2022-06-23
Add KLAUS_URL build argument to Dockerfile

s-ol:
This allows building forks etc by specifying e.g. a git+http:// URL.
If the argument is not specified, KLAUS_VERSION behaves as before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,4 +9,5 @@ RUN apk add --no-cache python3-dev py3-pip gcc musl-dev && \
     apk del python3-dev gcc musl-dev
 
 ARG KLAUS_VERSION
-RUN pip3 install klaus${KLAUS_VERSION:+==}${KLAUS_VERSION}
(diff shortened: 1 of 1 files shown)


jonashaag:
Hm I think both are legitimate. For people looking for an easy and reliable way to deploy Klaus the current version makes more sense. We could have two Dockerfiles.

s-ol:
I guess it's nice that the Dockerfile is "standalone" and can be copied out, but I'm not sure what the use case is: if you don't modify it you can run `docker build https://github.com/jonashaag/klaus.git` ([see here](https://docs.docker.com/engine/reference/commandline/build/#git-repositories)) or use `build: https://github.com/jonashaag/klaus.git` in docker compose and it will fetch any specified commit in the repo and build that exact version (or the latest).

jonashaag:
I agree, changed it in master.

PR #283, opened 2022-07-15
Fall back to just converting unknown bytestings using latin1, replacing invalid characters

jelmer:
Fall back to just converting unknown bytestings using latin1, replacing invalid characters.

This at least allows rendering pages rather than throwing 500s.

### DIFF
--- a/klaus/utils.py
+++ b/klaus/utils.py
@@ -144,7 +144,8 @@ def force_unicode(s):
     try:
         return s.decode("utf-8")
     except UnicodeDecodeError as exc:
-        last_exc = exc
+        pass
+
     try:
         return s.decode(locale.getpreferredencoding())
     except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


jonashaag:
WDYT about printing the exception or a warning in addition to that?

jelmer:
I don't think this is useful as an exception to log to whoever is running klaus - it's a problem in the data, not in the code.

Ideally we'd surface this in the web UI; something in red indicating that there were problems converting e.g. determining the encoding for a commit message.

jonashaag:
I don't agree that it wouldn't be useful for klaus administrators to have this kind of warning in the logs because the vast majority of them will also be a contributor to the repos they are hosting. But I just realized we don't have any infrastructure for logging so far, so let's just merge as-is. I guess it will be self-evident from the scrambled strings that something is wrong with the encoding.

PR #285, opened 2022-07-15
Support viewing blame of empty files

jelmer:
Support viewing blame of empty files.

Without this, klaus tries to access repo[b''] which leads to a KeyError.

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -14,5 +14,3 @@ if [ $# -eq 0 ]; then
 else
   pytest "$@"
 fi
-
-
--- a/runtests.sh
+++ b/runtests.sh
@@ -8,9 +8,11 @@
   done
 )
 
-export PYTHONPATH=tests
+export PYTHONPATH=tests:${PYTHONPATH}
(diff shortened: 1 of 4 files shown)


tests/test_blame.py
```diff
@@ -4,7 +4,13 @@
 
 def test_blame():
     with serve():
-        response = requests.get(UNAUTH_TEST_REPO_URL + "blob/HEAD/test.c")
+        response = requests.get(UNAUTH_TEST_REPO_URL + "blame/HEAD/test.c")
```
jonashaag, line 7: Good catch!

runtests.sh
```diff
@@ -8,9 +8,11 @@
...
 
-export PYTHONPATH=tests
+export PYTHONPATH=tests:${PYTHONPATH}
 if [ $# -eq 0 ]; then
   pytest -v tests
 else
   pytest "$@"
 fi
+
+
```
jonashaag, line 18:
```suggestion
```

jonashaag:
Thanks!

PR #289, opened 2022-08-06
Expose port 80 in Dockerfile

### LINKED ISSUE #288: Add EXPOSE to the Dockerfile
Synology DSM provides the possibility to expose a container through the web station, but for the feature to be enabled the dockerfile must include the EXPOSE instruction.

See also: https://kb.synology.com/en-us/DSM/help/Docker/docker_container?version=7

Could it be added?

I can provide a PR in case.

Thanks!

antoniomacri:
Fixes #288


jonashaag:
Actually I just realised that we don't use the port 80 anywhere in the Docker setup. So I would recommend creating your own Docker image instead:


```
FROM jonashaag/klaus
EXPOSE 80
```

antoniomacri:
You're right, I understand.
Feel free to close the issue.

Issue #293, opened 2022-09-02
Handle SymRefLoop


jelmer:
Broken repositories sometimes contain symrefs with loops in them. Klaus should handle the SymRefLoop exception and do something sensible (e.g. greying out the symref in lists and rendering an error).

I'm hoping to submit a PR to do this later this week.

PR #295, opened 2022-09-02
Handle SymrefLoop

jelmer:
Handle SymrefLoop. Fixes: #293

### DIFF
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -9,6 +9,14 @@ from dulwich.objects import Blob
 from dulwich.errors import NotTreeError
 import dulwich, dulwich.patch
 
+try:
+    from dulwich.refs import SymrefLoop
+except ImportError:  # dulwich < 0.20.46
+    InaccessibleRef = KeyError
+else:
+    InaccessibleRef = (SymrefLoop, KeyError)
+
+
 from klaus.utils import (
     force_unicode,
     parent_directory,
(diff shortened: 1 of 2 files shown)


klaus/repo.py
```diff
@@ -136,9 +142,14 @@ def get_default_branch(self):
...
             except KeyError:
                 pass
-        try:
-            return self.get_branch_names()[0]
-        except IndexError:
+            except SymrefLoop:
+                pass
+        for name in self.get_branch_names():
+            try:
+                self.get_commit(name)
```
jonashaag, line 149: Missing return

jonashaag:
Why aren’t actions running? :(

jonashaag:
Can you please rebase latest master

jonashaag:
WDYT about having the tuple `(KeyError, SymrefLoop)` defined as a constant somewhere and using that instead of repeating the list of exceptions everywhere? Could also have a short comment there what’s SymrefLoop.

jelmer:
> WDYT about having the tuple `(KeyError, SymrefLoop)` defined as a constant somewhere and using that instead of repeating the list of exceptions everywhere? Could also have a short comment there what’s SymrefLoop.

Happy to define a shorthand for those two places where we catch both (InaccessibleRef, perhaps?). That said, I'm wary of using that everywhere - e.g. in views.py where a message to the user saying a ref doesn't exist is quite different from saying the ref is broken.

jelmer:
Updated.

jonashaag:
Perfect!

PR #303, opened 2022-11-27
Add basic pyproject.toml, declaring that klaus uses setuptools

jelmer:
Add basic pyproject.toml, declaring that klaus uses setuptools


jonashaag:
Thanks!

PR #306, opened 2023-01-19
Replace invalid characters even when chardet has found an encoding

jelmer:
Replace invalid characters even when chardet has found an encoding

In some cases, it seems to find an encoding that's not valid for the entire string.


jonashaag:
Thanks!

PR #308, opened 2023-03-20
get_default_branch: Use HEAD instead of guessing

### LINKED ISSUE #307: Browsing to a repo shows the most recently updated branch, rather than the default
When I browse to a repo in Klaus whose default branch (as defined in `repo.git/HEAD`), if I have made a commit to a *different* branch (e.g. "foo") more recently than the most recent commit on HEAD, it will show me branch `foo` instead, which is unexpected and, in my opinion, undesirable.

mattcen:
Fixes #307

### DIFF
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -137,20 +137,13 @@ class FancyRepo(dulwich.repo.Repo):
         raise KeyError(rev)
 
     def get_default_branch(self):
-        """Tries to guess the default repo branch name."""
-        for candidate in ["master", "main", "trunk", "default", "gh-pages"]:
-            try:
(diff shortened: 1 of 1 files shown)


mattcen:
Not sure if this is t he nicest way to do things, especially with the `decode()` and calling `head()` just so we can catch it if it fails, but it feels less bad than guessing, which is what this replaces. Thoughts and comments welcome.

jelmer:
Rather than checking the revision that HEAD and other branches point at, you could simply check what HEAD resolves to.

See ``Repo.refs.follow(b'HEAD')``

mattcen:
@jelmer Thanks; code updated.

PR #310, opened 2023-04-05
Fix slow repo_list with many refs

### LINKED ISSUE #309: large repository makes index page very slow
# Issue

Similar to #230 I too have cloned the [nixpkgs](https://github.com/NixOS/nixpkgs/) repository and am now experiencing long wait times.
The corresponding path is quite slow, but it doesn't run into the default gunicorn timeout of 30s (though it's almost there with its load time of ~20s).
However when I access the index page – the one listing all repositories – I get a 5xx error from nginx due to a gateway timeout.

I opened this one with klaus rather than dulwich because it only seems to get triggered by the index page, *not* the project specific page (and the only other project there has like 100 commits and has like a megabyte I guess).

# Details

Commands to reproduce:

```bash
# go somewhere with ~5GiB of diskspace
cd "$(mktemp -dp /tmp)"
# this takes ages
git clone --mirror https://github.com/NixOS/nixpkgs
# it did happen with other command lines but this is the last one in my history that I know worked
klaus * --port 8888 --debug

# this should run pretty fast (0.395s)
curl http://localhost:8888/nixpkgs/
# this will take very long (35.289s)
curl http://localhost:8888/
```

I can reproduce this on both Gentoo and NixOS (klaus 2.0.2, dulwich 0.21.3 and 0.20.50 respectively).
I can also reproduce it with the Docker image (`dd9eaa5d40c7`) using this:

```bash
podman run --rm -p 7777:80 -v $PWD/:/repos/ -it jonashaag/klaus:latest klaus --host 0.0.0.0 --port 80 /repos/nixpkgs.git
```

Note that sometimes I also get a traceback like the following after a minute or so of dulwich reading the pack files, the error seems spurious, no idea what it's about, it usually goes away after one or two occurrences:

<details>
<summary>zlib.error: Error -3 while decompressing data: incorrect header check</summary>

```text
  File "/usr/lib/python3.10/site-packages/flask/app.py", line 2551, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/lib/python3.10/site-packages/klaus/utils.py", line 62, in __call__
    return super(ProxyFix, self).__call__(environ, start_response)
  File "/usr/lib/python3.10/site-packages/werkzeug/middleware/proxy_fix.py", line 187, in __call__
    return self.app(environ, start_response)
  File "/usr/lib/python3.10/site-packages/flask/app.py", line 2531, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/lib/python3.10/site-packages/flask/app.py", line 2528, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.10/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3.10/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/lib/python3.10/site-packages/klaus/views.py", line 79, in repo_list
    repos = sorted(repos, key=sort_key)
  File "/usr/lib/python3.10/site-packages/klaus/views.py", line 75, in <lambda>
    -(repo.fast_get_last_updated_at() or -1),
  File "/usr/lib/python3.10/site-packages/klaus/repo.py", line 363, in fast_get_last_updated_at
    self.__last_updated_at = self.__repo.get_last_updated_at()
  File "/usr/lib/python3.10/site-packages/klaus/repo.py", line 65, in get_last_updated_at
    return cached_call(
  File "/usr/lib/python3.10/site-packages/klaus/repo.py", line 36, in cached_call
    data = producer()
  File "/usr/lib/python3.10/site-packages/klaus/repo.py", line 68, in <lambda>
    producer=lambda: self._get_last_updated_at(all_refs),
  File "/usr/lib/python3.10/site-packages/klaus/repo.py", line 75, in _get_last_updated_at
    resolveable_refs.append(self[ref_hash])
  File "/usr/lib/python3.10/site-packages/dulwich/repo.py", line 771, in __getitem__
    return self.object_store[self.refs[name]]
  File "/usr/lib/python3.10/site-packages/dulwich/object_store.py", line 123, in __getitem__
    type_num, uncomp = self.get_raw(sha1)
  File "/usr/lib/python3.10/site-packages/dulwich/object_store.py", line 541, in get_raw
    return pack.get_raw(sha)
  File "/usr/lib/python3.10/site-packages/dulwich/pack.py", line 2377, in get_raw
    obj_type, obj = self.data.get_object_at(offset)
  File "/usr/lib/python3.10/site-packages/dulwich/pack.py", line 1328, in get_object_at
    unpacked = self.get_unpacked_object_at(offset, include_comp=False)
  File "/usr/lib/python3.10/site-packages/dulwich/pack.py", line 1313, in get_unpacked_object_at
    unpacked, _ = unpack_object(self._file.read, include_comp=include_comp)
  File "/usr/lib/python3.10/site-packages/dulwich/pack.py", line 871, in unpack_object
    unused = read_zlib_chunks(
  File "/usr/lib/python3.10/site-packages/dulwich/pack.py", line 297, in read_zlib_chunks
    decomp = decomp_obj.decompress(add)
zlib.error: Error -3 while decompressing data: incorrect header check
```

</details>

jonashaag:
@benaryorg please try this, closes #309

### DIFF
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -64,42 +64,35 @@ class FancyRepo(dulwich.repo.Repo):
         Cache is invalidated if one of the ref targets changes,
         eg. a new commit has been made and 'refs/heads/master' was changed.
         """
+        def _get_commit_time_cached(ref_id):
+            return cached_call(
+                key=(ref_id, "_get_commit_time"),
+                validator=None,
+                producer=lambda: _get_commit_time(ref_id),
+            )
+
+        def _get_commit_time(ref_id):
+            try:
+                return self[ref_id].commit_time
+            except (KeyError, AttributeError):
+                # Missing or non-commit object
+                return None
+
         max_refs = 1000
         if len(self.refs.keys()) > max_refs:
             # If we have too many refs, look at the branches only. (And HEAD, see below.)
             base = b"refs/heads"
         else:
             base = None
-        all_refs = self.refs.as_dict(base)
+        all_ids = list(self.refs.as_dict(base).values())
         # If we still have too many refs, keep only some.
-        if len(all_refs) > max_refs:
-            all_refs = dict(list(all_refs.items())[:max_refs])
+        if len(all_ids) > max_refs:
+            all_ids = sorted(all_ids)[:max_refs]
         # Always add HEAD.
-        all_refs[b"HEAD"] = self.refs[b"HEAD"]
+        all_ids.append(self.refs[b"HEAD"])
 
-        return cached_call(
-            key=(id(self), "get_last_updated_at"),
-            validator=all_refs,
-            producer=lambda: self._get_last_updated_at(all_refs.values()),
-        )
-
-    def _get_last_updated_at(self, all_refs):
-        resolveable_refs = []
-        for ref_hash in all_refs:
-            try:
-                resolveable_refs.append(self[ref_hash])
-            except KeyError:
-                # Whoops. The ref points at a non-existant object
-                pass
-        resolveable_refs.sort(
-            key=lambda obj: getattr(obj, "commit_time", float("-inf")), reverse=True
-        )
-        for ref in resolveable_refs:
-            # Find the latest ref that has a commit_time; tags do not
-            # have a commit time
-            if hasattr(ref, "commit_time"):
-                return ref.commit_time
-        return None
+        commit_times = filter(None, map(_get_commit_time_cached, all_ids))
+        return max(commit_times, default=None)
 
     @property
     def cloneurl(self):
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -92,7 +92,11 @@ class FancyRepo(dulwich.repo.Repo):
         all_ids.append(self.refs[b"HEAD"])
 
         commit_times = filter(None, map(_get_commit_time_cached, all_ids))
-        return max(commit_times, default=None)
+        try:
+            return max(commit_times)
+        except ValueError:
+            # Python 2 does not support max(..., default=)
+            return None
 
     @property
     def cloneurl(self):
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -58,10 +58,21 @@ class FancyRepo(dulwich.repo.Repo):
 
     # TODO: factor out stuff into dulwich
     def get_last_updated_at(self):
-        """Get datetime of last commit to this repository."""
-        # Cache result to speed up repo_list.html template.
-        # If self.get_refs() has changed, we should invalidate the cache.
-        all_refs = self.get_refs()
+        """Get datetime of last commit to this repository.
+
+        Caches the result to speed up the repo_list page.
+        Cache is invalidated if one of the ref targets changes,
+        eg. a new commit has been made and 'refs/heads/master' was changed.
+        """
+        if len(self.refs.keys()) > 10000:
+            # If we have too many refs, look at the branches only. (And HEAD, see below.)
+            base = b"refs/heads"
+        else:
+            base = None
+        all_refs = self.refs.as_dict(base)
+        # Always add HEAD.
+        all_refs[b"HEAD"] = self.refs[b"HEAD"]
+
         return cached_call(
             key=(id(self), "get_last_updated_at"),
             validator=all_refs,
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -64,19 +64,23 @@ class FancyRepo(dulwich.repo.Repo):
         Cache is invalidated if one of the ref targets changes,
         eg. a new commit has been made and 'refs/heads/master' was changed.
         """
-        if len(self.refs.keys()) > 10000:
+        max_refs = 1000
+        if len(self.refs.keys()) > max_refs:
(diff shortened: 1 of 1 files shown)


klaus/repo.py
```diff
@@ -58,10 +58,21 @@ def namespaced_name(self):
...
-        # Cache result to speed up repo_list.html template.
-        # If self.get_refs() has changed, we should invalidate the cache.
-        all_refs = self.get_refs()
+        """Get datetime of last commit to this repository.
+
+        Caches the result to speed up the repo_list page.
+        Cache is invalidated if one of the ref targets changes,
+        eg. a new commit has been made and 'refs/heads/master' was changed.
+        """
+        if len(self.refs.keys()) > 10000:
```
jelmer, line 67: 10000 still seems like quite a lot, there's no way to meaningfully display that - and it still means 10k random accesses of the repository.

What if there are more than 10000 heads, as is the case with the nix repo?

jonashaag: Ah damn I thought that most heads are branches but eg. GH PRs are also heads

jonashaag: Note that the number used here is not displayed anywhere, just the maximum timestamp of the refs

benaryorg:
> @benaryorg please try this

```text
curl -so /dev/null http://localhost:8000/  0.01s user 0.01s system 0% cpu 26.207 total max RSS 8
```

That's the initial request, still taking quite a long time (in fact almost hitting the default gunicorn timeout again).
After that of course it's cached and returns almost instantly.
The time I got with my quick&dirty version of the patch that only checked *HEAD* ran in ~25.514s which is barely any better, so I'd say this patch is as fast as this is going to get on that (admittedly slow) machine here.
So yeah, this solves my issue!
Thanks a lot!

---

For posterity's sake; the nixpkgs repository itself (which at this point already had its timestamp cached):

```text
curl -so /dev/null http://localhost:8000/nixpkgs/  0.01s user 0.01s system 0% cpu 2.610 total max RSS 8
```

klaus/repo.py
jonashaag: > Ah damn I thought that most heads are branches but eg. GH PRs are also heads

Wait no they are not

jonashaag:
Hmm I wonder why it's still so slow. For me it takes around 1s. Is the repo on a slow hard drive?

@jelmer is there a way to batch-lookup timestamps with Dulwich that's faster?

jelmer:
> Hmm I wonder why it's still so slow. For me it takes around 1s. Is the repo on a slow hard drive?
> 
> @jelmer is there a way to batch-lookup timestamps with Dulwich that's faster?

The main alternative is to iterate over most of the repository using something like Repo.object_store.iterobject_subset(). That will at least remove the need to reinflate delta bases more than once (which you would probably be doing when doing random access).

jonashaag:
```py
next(repo.object_store.iterobjects_subset([repo[b"HEAD"].id]))
```

```
KeyError: b'5804b7fb5fa07ac9b3720a19aa207edfb1521133'
```

Am I using it wrong?

jelmer:
> 



> ```python
> next(repo.object_store.iterobjects_subset([repo[b"HEAD"].id]))
> ```
> 
> ```
> KeyError: b'5804b7fb5fa07ac9b3720a19aa207edfb1521133'
> ```
> 
> Am I using it wrong?

That seems right, and works here (although on a different repo, obviously). Does repo[repo[b'HEAD'].id] work? What version of dulwich are you on?

benaryorg:
> Hmm I wonder why it's still so slow. For me it takes around 1s. Is the repo on a slow hard drive?

A slow SSD, which should still be a lot faster than an HDD but not up to speed with current SSDs.
However these load times aren't IO bound but CPU bound.
The thread the request is running on *does* use up 100% of a single core on the machine during those 20s though, so it likely isn't impacted by any IO speeds.
Recalling the stacktrace I got at one point I'm pretty sure it's going through a lot of of zlib compressed data, though I can't say for sure what exactly takes so long.

jonashaag:
> Repo.object_store.iterobject_subset()

Unfortunately it is actually slower:

```
In [11]: %timeit list(repo.object_store.iterobjects_subset(random.sample(list(refs.values()), 1000), allow_missing=True))
94.2 ms ± 2.46 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [12]: %timeit [repo[x] for x in random.sample(list(refs.values()), 1000)]
66.7 ms ± 2.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

jonashaag:
@benaryorg what's your CPU? I wonder how it can be 20x slower than mine

benaryorg:
> @benaryorg what's your CPU? I wonder how it can be 20x slower than mine

Intel(R) Atom(TM) CPU N2800   @ 1.86GHz

This is probably one of my oldest and slowest servers *\*sigh\**

PR #314, opened 2023-06-03
Fix some issues reported by ruff

jelmer:
In particular:
 * use def rather than assigning lambda function
 * unused imports
 * 'X not in Y' vs 'not X in Y'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ tests/repos/build
 build/
 dist/
 .DS_Store
+.mypy_cache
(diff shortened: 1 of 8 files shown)


jonashaag:
What’s your option on the lambda vs def rule? I don’t like it but have no strong opinion. 

If you want to add ruff, black, etc to the CI feel free to!

jelmer:
> What’s your option on the lambda vs def rule? I don’t like it but have no strong opinion.

I don't have a particularly strong opinion either way, but since other tooling (ruff, pycodestyle) seems to prefer def over assigning lambdas I'm inclined to just go with that.

PR #327, opened 2024-05-13
Maybe fix #326

### LINKED ISSUE #326: AttributeError: 'FancyRepo' object has no attribute 'object_store'
Here is an error from trying to clone a repository using klaus:
```
  File "/usr/local/lib/python3.9/site-packages/gunicorn/workers/sync.py", line 183, in handle_request
    for item in respiter:
  File "/usr/local/lib64/python3.9/site-packages/dulwich/web.py", line 231, in get_info_refs
    handler.handle()
  File "/usr/local/lib64/python3.9/site-packages/dulwich/server.py", line 379, in handle
    self.repo.object_store,
AttributeError: 'FancyRepo' object has no attribute 'object_store'
```
Python version 3.9.18 and klaus installed using pip. Here is how klaus was instantiated:
```
/usr/local/bin/klaus --smarthttp /var/lib/git/repo.git/
```
It seems object_store, self.repo.get_peeled, and self.repo.refs.get_symrefs are all undefined.  This is on Rocky Linux 9.3.

jonashaag:
@rbeester

### DIFF
--- a/klaus/repo.py
+++ b/klaus/repo.py
@@ -60,6 +60,14 @@ class FancyRepo:
         return self.dulwich_repo.path
 
     @property
+    def object_store(self):
+        return self.dulwich_repo.object_store
+
+    @synchronized
+    def __getitem__(self, key):
+        return self.dulwich_repo[key]
+
+    @property
     def name(self):
         return repo_human_name(self.path)
 
@@ -69,10 +77,6 @@ class FancyRepo:
             return f"~{self.namespace}/{self.name}"
         else:
             return self.name
-
-    @synchronized
-    def __getitem__(self, key):
-        return self.dulwich_repo[key]
 
     # TODO: factor out stuff into dulwich
     @synchronized


rbeerster:
Ok, that is a start.  However, as I mentioned there are other attributes not defined.  Using your change, here I what I came up with.  Definitely look this over carefully.  Just because it worked does not mean it is correct.  Here is a diff:
```
diff -u /usr/local/lib/python3.9/site-packages/klaus/repo.py{.20240513,}
--- /usr/local/lib/python3.9/site-packages/klaus/repo.py.20240513	2024-05-13 19:35:41.427882686 -0400
+++ /usr/local/lib/python3.9/site-packages/klaus/repo.py	2024-05-13 20:00:06.162565145 -0400
@@ -58,6 +58,26 @@
     @property
     def path(self):
         return self.dulwich_repo.path
+    
+    @property
+    def object_store(self):
+        return self.dulwich_repo.object_store
+    
+    @property
+    def get_peeled(self):
+        return self.dulwich_repo.get_peeled
+    
+    @property
+    def refs(self):
+        return self.dulwich_repo.refs
+    
+    @property
+    def find_missing_objects(self):
+        return self.dulwich_repo.find_missing_objects
+    
+    @synchronized
+    def __getitem__(self, key):
+        return self.dulwich_repo[key]
 
     @property
     def name(self):
@@ -70,10 +90,6 @@
         else:
             return self.name
 
-    @synchronized
-    def __getitem__(self, key):
-        return self.dulwich_repo[key]
-
     # TODO: factor out stuff into dulwich
     @synchronized
     def get_last_updated_at(self):
```

jonashaag:
Ok that's terrible, I'll have to proxy all of the attributes with a catch-all solution or maybe we can pass the inner actual repo object around instead of the wrapper.

jonashaag:
I don't understand why this isn't covered in our tests though...

PR #334, opened 2024-07-14
python 3.12 support

fin444:
remove references to distutils
- `app_args.py`: add polyfill
- `setup.py`: remove code that hasn't been necessary since 51094d7

### DIFF
--- a/klaus/contrib/app_args.py
+++ b/klaus/contrib/app_args.py
@@ -1,13 +1,15 @@
 import os
 
+
 def strtobool(val):
     val = val.lower()
-    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+    if val in ("y", "yes", "t", "true", "on", "1"):
         return 1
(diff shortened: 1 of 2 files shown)


jonashaag:
Thanks for the patch! Do you think you know what you're doing here? Because I don't when it comes to these packaging things and have no way of verifying because I have no clue what's right and wrong. So I'll have to trust your judgment here.

jelmer:
FWIW this looks good to me.

fin444:
I'm not exactly an expert on this, but for `app_args.py` it's just copy pasting the function from the library. For `setup.py`, what I found suggested that the hack was only necessary if `data_files` was included as a parameter ([source 1](https://stackoverflow.com/a/3042436), [source 2](https://github.com/celery/kombu/pull/1467#pullrequestreview-849763856), source 3 it just says it in the comment)

fin444:
And while clean up would be cool, I don't know much about packaging lol. I only submitted this because NixOS updated to 3.12 and it broke my stuff

Issue #337, opened 2025-07-16
More reference docs


onel:
I recently discovered this project and I wanted to ask if you're open to accepting some contribution regarding the docs?
I see most of the code is well documented but some parts are missing.

jonashaag:
yes absolutely 

maybe we can also ask an LLM to document some things

onel:
Yeah, definitely. I'll go over the parts of the code base that need more it and push some updates soon.

PR #338, opened 2025-07-28
Added more reference docs

onel:
This is a first round of adding minimal reference docs.
Working on a more comprehensive addition right now.

Issue: #337


setup.py
```diff
@@ -2,8 +2,10 @@
 
 from setuptools import setup
 
+# Read the long description from the README file
```
jelmer: Is this really adding anything? This is just describing what the code is actually doing.

klaus/ctagsutils.py
```diff
@@ -19,7 +23,12 @@ def create_tagsfile(git_repo_path, git_rev):
     This creates a temporary clone of the repository, checks out the revision,
     runs 'ctags -R' and deletes the temporary clone.
 
-    :return: path to the generated tagsfile
+    Args:
+        git_repo_path (str): Path to the Git repository.
```
jelmer: wouldn't it be better to add typing hints, rather than documenting types in the docstring?

setup.py
onel: It's usually a good idea to also have a plain english explanation of the code even if trivial, becaue AI code assistants and embedding models understand english better. It makes finding these blocks of code faster. 
Down to remove it if it's not wanted but I would recommend adding them

klaus/ctagsutils.py
onel: 100%. I mentioned to @jonashaag that I would only add documentation so I don't want to expand the scope of this PR too much.
I also think code in general should have both, no? They serve different purposes.
Do doc site generation tools also look at typing or they do they only follow docstrings?

setup.py
jelmer: The primary consumer of code should still be humans, not AIs. Also, if a LLM really needs this comment it's not very good - this is fairly trivial and standard.

This is the equivalent of:

```
# print: blah
print("blah")
```

klaus/ctagsutils.py
jelmer: Yes, most look at type hints. Adding type info in the docstrings first and then moving them to type hints seems like unnecessary churn - but it's obviously up to @jonashaag what he prefers here.

klaus/contrib/app_args.py
```diff
@@ -2,6 +2,21 @@
...
+
+    Returns:
+        int: 1 for truthy values, 0 for falsy values.
+
+    Raises:
+        ValueError: If the input string is not a recognized truth value.
+
+    Note:
+        Truthy values: 'y', 'yes', 't', 'true', 'on', '1' (case insensitive)
+        Falsy values: 'n', 'no', 'f', 'false', 'off', '0' (case insensitive)
```
jonashaag:
```suggestion
```

Please don't add comments like these. They don't add anything useful. Everything in this docstring is already in the code. The only potentially valuable thing here is the type of the parameter; please add a type hint instead.

klaus/contrib/app_args.py
```diff
@@ -12,6 +27,23 @@ def strtobool(val):
...
+    """Extract Klaus application arguments and configuration from environment variables.
+
+    Returns:
+        tuple: A tuple containing (args, kwargs) where:
+            - args (tuple): Positional arguments (repos list, site name)
+            - kwargs (dict): Keyword arguments with configuration options
+
+    Environment Variables:
+        KLAUS_REPOS: Space-separated list of repository paths
+        KLAUS_SITE_NAME: Name of the site (default: "unnamed site")
```
jonashaag: Please don't add default values here. They are very obvious from reading the code.

klaus/contrib/app_args.py
```diff
@@ -12,6 +27,23 @@ def strtobool(val):
 
 
 def get_args_from_env():
+    """Extract Klaus application arguments and configuration from environment variables.
+
+    Returns:
```
jonashaag: Please make these type hints

klaus/contrib/wsgi.py
```diff
@@ -1,9 +1,18 @@
...
+
+This module creates a WSGI application instance by reading configuration
+from environment variables and setting up optional HTTP digest authentication.
+"""
+
 from klaus import make_app
 
 from .app_args import get_args_from_env
 
+# Get application arguments and keyword arguments from environment variables
```
jonashaag: This is useless.

klaus/contrib/wsgi_autoreload.py
```diff
@@ -1,24 +1,41 @@
...
     warnings.warn(
         "use KLAUS_REPOS_ROOT instead of KLAUS_REPOS for the autoreloader apps",
         DeprecationWarning,
     )
 
+# Get application arguments and keyword arguments from environment variables
 args, kwargs = get_args_from_env()
+
+# Determine the repository root path from environment variables
+# Prefer KLAUS_REPOS_ROOT over the deprecated KLAUS_REPOS
```
jonashaag: This is useless.

klaus/contrib/wsgi_autoreload.py
```diff
@@ -1,24 +1,41 @@
...
 repos_root = os.environ.get("KLAUS_REPOS_ROOT") or os.environ["KLAUS_REPOS"]
+
+# Replace the first argument with the repository root path
```
jonashaag: This is useless.

klaus/contrib/wsgi_autoreload.py
```diff
@@ -1,24 +1,41 @@
...
 args = (repos_root,) + args[1:]
 
+# Handle htdigest file caching if specified
 if kwargs["htdigest_file"]:
     # Cache the contents of the htdigest file, the application will not read
     # the file like object until later when called.
     with open(kwargs["htdigest_file"], encoding="utf-8") as htdigest_file:
         kwargs["htdigest_file"] = io.StringIO(htdigest_file.read())
 
+# Create the auto-reloading WSGI application with processed arguments
```
jonashaag: This is useless.

klaus/ctagsutils.py
```diff
@@ -41,5 +50,9 @@ def create_tagsfile(git_repo_path, git_rev):
 
 
 def delete_tagsfile(tagsfile_path):
-    """Delete a tagsfile."""
+    """Delete a tagsfile.
```
jonashaag: This entire docstring is useless, maybe modulo the type (which should be a type hint)

tools/dumbtest.py
```diff
@@ -9,6 +9,14 @@
 
 
 def view_from_url(url):
+    """Extract the domain/host portion from a URL.
```
jonashaag: I think this is wrong?

jonashaag: Also this is a testing script, not sure if it's worth documenting anything here

setup.py
jonashaag: Sorry but all comments made in this file are useless.

jonashaag:
@onel thanks for taking a stab at this! Good comments should focus on the "why", not the "what" or the "how". The latter are already described by the code, and if you duplicate that information as comments, it gets out of sync with the implementation really quickly.

onel:
> @onel thanks for taking a stab at this! Good comments should focus on the "why", not the "what" or the "how". The latter are already described by the code, and if you duplicate that information as comments, it gets out of sync with the implementation really quickly.

Those are good points. Going through your comments and will update the PR soon

PR #339, opened 2025-07-29
Drop support for ancient dulwich versions

jelmer:


jonashaag:
Hm, CI failures seem unrelated but would be nice to have green CI.

I haven't really used Klaus for a while so I'm currently not going to invest time into fixing it though

jonashaag:
oh nice <3

jelmer:
On Thu, Jul 31, 2025 at 11:50:02PM -0700, Jonas Haag wrote:
>jonashaag left a comment (jonashaag/klaus#339)
>
>oh nice <3

So this actually helped find a regression in tree_lookup_path() - it
broke for "" - which I've fixed in the latest version of Dulwich.

Jelmer

